PR #10, opened 2023-05-20
Improved version of the "Forget" menu

Firiman:
Things I'm planning to add next:
*Confirmation window before forgetting a character
*Sorting
*Ability to start encounter with npc if in appropriate location
*Maybe some other stats to show


Game/GameUI.gd
```diff
@@ -33,6 +33,7 @@ onready var devCommentaryPanel = $HBoxContainer/DevCommentary
 var textboxes: Dictionary = {}
 var gameParser: GameParser
 var sayParser: SayParser
+onready var npcListScreen = $HBoxContainer/VBoxContainer2/NpcList
```
Alexofp, line 36: Making it so this screen is always 'loaded' is kinda bad because it's only used in one scene.

Take a look at the StrugglingScene.gd where I load a minigame screen dynamically
it would basically look like
```
var npclistScene = preload("res://UI/NpcList/NpcList.tscn")
....
(inside the nursery scene)
var npclist = npclistScene.instance()
GM.ui.addCustomControl("npclist", npclist)
```

Game/GameUI.gd
```diff
@@ -312,6 +314,24 @@ func getCurrentLocationName():
 	return mapAndTimePanel.getLocationName()
 
 
+func showNpcList():
```
Alexofp, line 317: You wouldn't need these functions either

NPCListButton.tres
Alexofp, line 1: These two .tres files shouldn't be in the root of the project. And I think you can just keep them embeded into the tscn files that they are from?

Scenes/EncountersMenuScene.gd
```diff
@@ -77,9 +79,9 @@ func _run():
 		addButton("Back", "Close this menu", "endthescene")
 		
 		if(hasSomeoneToForget):
-			addButton("Forget", "Pick which character you wanna forget", "forgetmenu")
+			addButton("NPC list", "Shows NPCs that you encountered", "npclistmenu")
```
Alexofp, line 82: See how I carefully avoided using the word NPC in my code? x3 They're characters, npc is too game-y

I understand that 'Character list' is too long. Maybe just rename it to 'Memory' then and change all 'NPC' to 'character'

Scenes/EncountersMenuScene.gd
```diff
@@ -165,48 +167,46 @@ func _run():
...
+#			if(desc == null):
+#				desc = ""
+
+#			var kidsAmount = GM.CS.getChildrenAmountOf(characterID)
+#			if(kidsAmount > 0):
+#				if(desc != ""):
+#					desc += "\n"
+
+			var NPCname = dynamicCharacter.getName()
+			var gender = dynamicCharacter.npcGeneratedGender.capitalize() #accessing param like this is kinda wrong but getGender() returns number and I don't want that
```
Alexofp, line 201: var gender = NpcGender.getVisibleName(dynamicCharacter.npcGeneratedGender)

Scenes/EncountersMenuScene.gd
```diff
@@ -165,48 +167,46 @@ func _run():
...
+			var subbyStat = dynamicCharacter.getPersonality().getStat(PersonalityStat.Subby)
+			var personality = PersonalityStat.getVisibleDesc("Subby", subbyStat)
```
Alexofp, line 203: var personality = PersonalityStat.getVisibleDesc(PersonalityStat.Subby, subbyStat)

Scenes/EncountersMenuScene.gd
```diff
@@ -165,48 +167,46 @@ func _run():
...
 			var sharedKidsAmount = GM.CS.getSharedChildrenAmount("pc", characterID)
-			if(sharedKidsAmount > 0):
-				if(sharedKidsAmount == kidsAmount):
-					desc += " All of them are from you."
-				else:
-					desc += " "+str(sharedKidsAmount)+" of them are from you."
+
+			GM.ui.npcListScreen.addRow(NPCname, gender, personality, characterID, sharedKidsAmount)
```
Alexofp, line 206: this would become 
`npclist.addRow(NPCname, gender, personality, characterID, sharedKidsAmount)`

Scenes/EncountersMenuScene.gd
```diff
@@ -165,48 +167,46 @@ func _run():
...
 			
-			addButton(dynamicCharacter.getName(), desc, "forget", [dynamicCharacter.getID()])
+		setState("npclistmenu")
+		_run()
```
Alexofp, line 209: why do this here? You're already inside the _run() function

Alexofp (commented):
After a quick glance, it would be good to change this. I might spot more things later. If you have any questions/problems with how to make the panel show up inside the scene without changing the GameUI, poke me ^^

also, if changing ui its better to attach a screenshot ^^. I know you showed it to me already, just for the future

Scenes/EncountersMenuScene.gd
Firiman: I made this to update state manually. setState() doesn't do this on its own.

Alexofp: Honestly, just copy-paste the code from the 'npclistmenu' state, it's fine. Because doing it like this might break at any point if I change the API

NPCListButton.tres
Firiman: Agree I'll move them to the npclist folder.

Game/GameUI.gd
Firiman: Turned to be a really bad advice because every button pressed triggers run() function inside SceneBase which has GM.ui.clearUItextboxes() call inside it so my loaded scene goes to null. And I need to select occupation with buttons.

UI/NpcList/NPCRow.gd
```diff
@@ -0,0 +1,54 @@
...
+func setPersonality(personality: String):
+	_npcPersonalityLabel.text = personality
+
+
+func setChildren(children):
+	_amountOfChildrenLabel.text = String(children)
+
+
+func setNpcID(ID: String):
+	if(_npcID == null):
```
Alexofp, line 30: this probably should be `if(ID == null):` right?
But when you specify the type of the argument Godot actually ensures that it won't be null so you don't really need that check. You would need it if your function declaration looked like `func setNpcID(ID):` (without the : String part)

UI/NpcList/NPCRow.gd
```diff
@@ -0,0 +1,54 @@
...
+func setGender(gender: String):
+	_npcGenderLabel.text = gender
+
+
+func setPersonality(personality: String):
+	_npcPersonalityLabel.text = personality
+
+
+func setChildren(children):
+	_amountOfChildrenLabel.text = String(children)
```
Alexofp, line 26: _amountOfChildrenLabel.text = str(children)

UI/NpcList/NPCRow.gd
```diff
@@ -0,0 +1,54 @@
...
+	else: 
+		push_error("Exception: NPCRow: NPC ID is missing")
+
+
+func _on_Forget_pressed():
+	if(_npcID == null):
+		push_error("Exception: Attempt to delete null NPC in the NPC list")
+	else:
+		var npcListObj = load("res://UI/NpcList/NpcList.gd").new()
+		npcListObj.forgetNPC(_npcID)
```
Alexofp, line 48: this looks very strange and I think you're leaking memory x3. You can either move the forgetNPC function into this class or create a signal and subscribe to it inside the NpcList. Signals are like events basically, perfect for this. Here is an example https://github.com/Alexofp/BDCC/blob/main/UI/SkillsUI/PerkButton.gd (look for the signal and how its being used)

you basically create a signal in this class `signal onForgetButtonPressed(npcID)`
then you emit it in this function `emit_signal("onForgetButtonPressed", _npcID)`
And then inside the NpcList class you subscribe to it dynamically while creating the rows `var _ok = someNpcRow.connect("onForgetButtonPressed", self, "forgetNPC")`
should work ^^

UI/NpcList/NpcList.gd
```diff
@@ -0,0 +1,26 @@
+extends Control
+
+onready var npcRow = load("res://UI/NpcList/NPCRow.tscn")
+onready var container = get_node("VBoxContainer/ItemList/NpcListControld/ScrollContainer/HBoxContainer")
```
Alexofp, line 4: onready var container = $VBoxContainer/ItemList/NpcListControld/ScrollContainer/HBoxContainer

UI/NpcList/NpcList.gd
```diff
@@ -0,0 +1,26 @@
...
+
+
+func addRow(name: String, gender: String, personality: String, ID, children: int = 0):
+	var newRow = npcRow.instance()
+	container.add_child(newRow)
+	newRow.setName(name)
+	newRow.setGender(gender)
+	newRow.setPersonality(personality)
+	newRow.setNpcID(ID)
+	newRow.setChildren(children)
```
Alexofp, line 14: Yeah, you just connect to the signal in this function ^^

UI/NpcList/NPCRow.gd
```diff
@@ -0,0 +1,54 @@
...
+
+func setNpcID(ID: String):
+	if(_npcID == null):
+		push_error("Exception: Attempt to set null NPC in the NPC list")
+	else:
+		_npcID = ID
+
+
+func getNpcID():
+	if(_npcID != null):
```
Alexofp, line 37: since you declared it as `_npcID: String` it will never be null ^^ (I think at least x3). You can still check it against an empty string

UI/NpcList/NPCRow.gd
```diff
@@ -0,0 +1,54 @@
...
+		return _npcID
+	else: 
+		push_error("Exception: NPCRow: NPC ID is missing")
```
Alexofp, line 40: use Log.error() instead, it will show up in the game's console then

UI/NpcList/NPCRow.gd
```diff
@@ -0,0 +1,54 @@
...
+
+func setGender(gender: String):
+	_npcGenderLabel.text = gender
+
+
+func setPersonality(personality: String):
+	_npcPersonalityLabel.text = personality
+
+
+func setChildren(children):
```
Alexofp, line 25: it should probably be setChildrenAmount right?

Scenes/EncountersMenuScene.gd
```diff
@@ -1,16 +1,20 @@
 extends SceneBase
 
-var pickedPoolToForget = ""
+var pickedPoolToShow = ""
 var pickedFetishToChange = ""
 var pickedPersonalityStat = ""
 var pickedGenderToChange = ""
 var pickedSpeciesToChange = ""
+var npclistScene = preload("res://UI/NpcList/NpcList.tscn")
+var npclist
```
Alexofp, line 9: I don't think you need this variable to be here, it seems to be fine to declare it while creating the list inside the _run function `var npclist = npclistScene.instance()`

Alexofp (commented):
Looks much better, thank you ^^. I just have some little things you can still fix

UI/NpcList/NpcList.gd
Firiman: It's the same basically just a bit shorter? But ok.

UI/NpcList/NPCRow.gd
Firiman: push_error( ) "Pushes an error message to Godot's built-in debugger and to the **OS terminal.**" That's probably better?

Alexofp: Log.error is my function, it allows me to change the logging at any time without going through the whole codebase. Better to use the same function everywhere ^^

UI/NpcList/NpcList.gd
Alexofp: it works the same but I use $ everywhere. It's shorter too yeah ^^

UI/NpcList/NPCRow.gd
Firiman: Yes I wanted to make it with signals originally but creating an object **only** when you actually want to delete character seemed like a better choice than connecting **every** row.

Alexofp:
Thank you very much, I will credit you in the credits ^^

PR #11, opened 2023-05-23
Npclist sorting

Firiman:
Added sorting + minor refactoring of the previous code.

### DIFF
--- a/UI/NpcList/NPCRow.gd
+++ b/UI/NpcList/NPCRow.gd
@@ -8,23 +8,15 @@ onready var meetButton = $HBoxContainer/CenterContainer2/Meet
 onready var forgetButton = $HBoxContainer/CenterContainer/Forget
 onready var showNpcButton = $ShowNPC
 signal onForgetButtonPressed(npcID)
-var _npcID: String
+var _npcID setget setNpcID, getNpcID
 
 
-func setName(name: String):
+func init(name, gender, personality, ID, children):
 	_npcNameLabel.text = name
-
-
-func setGender(gender: String):
-	_npcGenderLabel.text = gender
-
-
-func setPersonality(personality: String):
-	_npcPersonalityLabel.text = personality
-
-
-func setChildrenAmount(children):
-	_amountOfChildrenLabel.text = str(children)
+	_npcGenderLabel.text  = gender
+	_npcPersonalityLabel.text  = personality
+	self._npcID = ID
+	_amountOfChildrenLabel.text  = str(children)
 
 
 func setNpcID(ID: String):
--- a/UI/NpcList/NPCRow.gd
+++ b/UI/NpcList/NPCRow.gd
@@ -10,14 +10,15 @@ onready var showNpcButton = $ShowNPC
 signal onForgetButtonPressed(npcID)
 var _npcID setget setNpcID, getNpcID
 var _npcSubbyStatRaw: float
+var _amountOfChildrenRaw: int
 
-
-func init(name, gender, subbyStat, ID, children):
+func initData(name, gender, subbyStat, ID, children):
 	_npcNameLabel.text = name
 	_npcGenderLabel.text  = gender
 	_npcPersonalityLabel.text  = PersonalityStat.getVisibleDesc(PersonalityStat.Subby, subbyStat)
 	_npcSubbyStatRaw = subbyStat
 	self._npcID = ID
+	_amountOfChildrenRaw = children
 	_amountOfChildrenLabel.text  = str(children)
 
 
@@ -48,7 +49,7 @@ func getSubbyStatRaw() -> float:
 
 
 func getAmountOfChildren() -> int:
-	return int(_amountOfChildrenLabel.text)
+	return _amountOfChildrenRaw
 
 
 func _on_Forget_pressed():
--- a/UI/NpcList/NPCRow.gd
+++ b/UI/NpcList/NPCRow.gd
@@ -1,0 +1,53 @@
+extends PanelContainer
+
+onready var _npcNameLabel = $HBoxContainer/Name
+onready var _npcGenderLabel = $HBoxContainer/Gender
+onready var _npcPersonalityLabel = $HBoxContainer/Personality
+onready var _amountOfChildrenLabel = $HBoxContainer/Children
+onready var meetButton = $HBoxContainer/CenterContainer2/Meet
+onready var forgetButton = $HBoxContainer/CenterContainer/Forget
+onready var showNpcButton = $ShowNPC
+signal onForgetButtonPressed(npcID)
+var _npcID: String
+
+
+func setName(name: String):
+	_npcNameLabel.text = name
+
+
+func setGender(gender: String):
+	_npcGenderLabel.text = gender
+
+
+func setPersonality(personality: String):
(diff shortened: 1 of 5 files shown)


UI/NpcList/NpcList.gd
```diff
@@ -1,27 +1,124 @@
...
+	_childrenBtnState = !_childrenBtnState
+
+
+func sortChildrenAmountAscending(a: Node, b: Node):
+	return a.getAmountOfChildren() < b.getAmountOfChildren()
+
+
+func _notification(what):
+	if(what == NOTIFICATION_PREDELETE && GM.main != null):
+		GM.main.playAnimation(StageScene.Solo, "stand", {npc=GM.pc})
```
Alexofp, line 124: This is way too complicated, just make it so the animation is played when you back out of the menu (make the back button look like `addButton("Back", "Go back a level", "closenpclist")` and then add a new check in the _react function)
```
if(_action == "closenpclist"):
	setState("")
	GM.main.playAnimation(StageScene.Solo, "stand")
	return
```

UI/NpcList/NpcList.gd
```diff
@@ -1,27 +1,124 @@
...
-func addRow(name: String, gender: String, personality: String, ID, children: int = 0):
+func addRow(name: String, gender: String, subbyStat: float, ID: String, children: int = 0):
 	var newRow = npcRow.instance()
 	container.add_child(newRow)
-	newRow.setName(name)
-	newRow.setGender(gender)
-	newRow.setPersonality(personality)
-	newRow.setNpcID(ID)
-	newRow.setChildrenAmount(children)
+	newRow.init(name, gender, subbyStat, ID, children)
```
Alexofp, line 18: I would rename the function to setData, init is throwing me off because `_init` is built-in godot's function

UI/NpcList/NPCRow.gd
```diff
@@ -41,6 +35,22 @@ func getNpcID():
...
+func getNpcGender():
+	return _npcGenderLabel.text
+
+
+func getSubbyStatRaw() -> float:
+	return _npcSubbyStatRaw
+
+
+func getAmountOfChildren() -> int:
+	return int(_amountOfChildrenLabel.text)
```
Alexofp, line 51: You can just create an int variable and store this so you don't have to cast strings to ints many times (and sorting something will do that -many- times x3)

UI/NpcList/NpcList.gd
```diff
@@ -1,27 +1,124 @@
...
+	personalityButton.pressed = false
+	childrenButton.pressed = false
+	
+	var nodesSortedArr = container.get_children()
+	
+	if(_nameBtnState):
+		nodesSortedArr.sort_custom(self, "sortNameAscending")
+	else:
+		nodesSortedArr.sort_custom(self, "sortNameAscending")
+		nodesSortedArr.invert()
```
Alexofp, line 38: I think you can just create a `sortNameDescending` function that's the opposite of the first one, means you will move less data around

UI/NpcList/NpcList.gd
```diff
@@ -1,27 +1,124 @@
...
+		
+	for nodeNum in nodesSortedArr.size():
+		container.move_child(nodesSortedArr[nodeNum], nodeNum)
```
Alexofp, line 41: Are you sure this works? x3 My head hurts trying to understand it but if it actually sorts it correctly then sure. I usually just remove all the childs and re-create them in the correct order which is way less fast but less error-prone x3

UI/NpcList/NpcList.gd
```diff
@@ -1,27 +1,124 @@
...
+	if(_personalityBtnState):
+		nodesSortedArr.sort_custom(self, "sortPersonalityAscending")
+	else:
+		nodesSortedArr.sort_custom(self, "sortPersonalityAscending")
+		nodesSortedArr.invert()
+		
+	for nodeNum in nodesSortedArr.size():
+		container.move_child(nodesSortedArr[nodeNum], nodeNum)
+	
+	_personalityBtnState = !_personalityBtnState
```
Alexofp, line 91: I see a lot of code duplication.

it can probably be simplified to something like:
```
func _on_Personality_pressed():
	unpressAllButtons()
	personalityButton.pressed = true

	if(_personalityBtnState):
		sortNpcList("sortPersonalityAscending")
	else:
		sortNpcList("sortPersonalityDescending")

	_personalityBtnState = !_personalityBtnState
```

Alexofp (commented):
Here is my feedback ^^. Not sure if this menu needs sorting but having it won't hurt, heh

UI/NpcList/NpcList.gd
Firiman: Works fine

UI/NpcList/NpcList.gd
Firiman: Looks complicated but it is a simple destructor/finalizer, works fine

Alexofp: I'd rather you change it still

Alexofp: godot doesn't really have destructors, what you're using here is a very low level function. If you don't like my way, I'd like to hear why ^^. Putting the logic inside the scene instead of UI is better, ideally UI shouldn't even know about GM.main but that's hard to achieve in all cases. Still, better to try ^^

Firiman: Honestly I do not like your buttons, every click is a roller-coaster through the half of game code, everything is recalculated like buffs etc I saw in the debugger. It belongs here as an npclist function, works well. GM.main needed to check if the game is still running, I can put anything else here to check, I just don't know what. So that part could be modified.

Firiman: Actually it works even without GM.main check, but throws error in the output window if you just close the game with that menu open and I didn't like that.

Alexofp: if you don't like how the buttons work you will have a bad time in the future because the whole game runs on those buttons x3. I tried to only recalculate buffs/status effects only when they need to, trust me. It never worked well, sometimes calculating all that stuff more than once, sometimes not recalculating it when needed. So the best approach that I found is.. doing it once every time the user picks an option, works in all cases even if not the most optimized solution (fine for a text game x3)

But yeah, then I suggest just removing it. It's fine if the animation doesn't resest, it will do that the next time the player moves or something. As you can see you run into problems while using this notification function, who knows how else it's gonna break

Firiman: I didn't run into any actual problems with notification and it is secluded anyway, only possible to be executed while in the characters menu. And it is actually called "destructor" in the official Godot documentation: 
destructor

Alexofp: You can't imagine how many unique godot quirks I've ran into x3

I trust you, your code works, I'm just trying to make it cleaner ^^

At the end of the day someone gotta have the final say, otherwise we will never stop arguing. Do you wanna have the final say in this project? x3

Alexofp:
Thank you ^^

UI/NpcList/NpcList.gd
Firiman: And I thank you for that! 90% of times I fully agree on suggested changes.

PR #14, opened 2023-05-26
Pregnancy (Fertility) perk tree

Firiman:


Skills/Perk/PregnancyBetterOvulation.gd
```diff
@@ -0,0 +1,25 @@
...
+func _init():
+	id = Perk.PregnancyBetterOvulation
+	skillGroup = Skill.Pregnancy
+
+func getVisibleName():
+	return "Better ovulation"
+
+func getVisibleDescription():
+	return "Your body is adapting to produce more offspring. Amount of the egg cells you release during each ovulation phase is increased."
+
```
Firiman, line 12: 250% max eggs total.

Skills/Perk/PregnancyBetterOvulationV2.gd
```diff
@@ -0,0 +1,28 @@
...
+	id = Perk.PregnancyBetterOvulationV2
+	skillGroup = Skill.Pregnancy
+
+func getVisibleName():
+	return "Breeding overdrive"
+
+func getVisibleDescription():
+	return "Your reproductive organs are pushing themselves to the limit for you to bear as many children as possible. \n" \
+			+ "Maximum of the egg cells you release during each ovulation phase is increased even further and fertility is improved by 20%"
+
```
Firiman, line 13: 450% max eggs total with the previous perk.

Skills/Perk/PregnancyMotherOfTheYear.gd
```diff
@@ -0,0 +1,24 @@
...
+	id = Perk.PregnancyMotherOfTheYear
+	skillGroup = Skill.Pregnancy
+
+func getVisibleName():
+	return "Mother of the year"
+
+func getVisibleDescription():
+	return "AlphaCorp can't thank you enough for providing more cannon fodder for their global operations. Oh! we meant to say what a nice kids you have we always happy to help them grow and prosper, right! \n" \
+			+ "You get a stats boost buff for 2 days after giving birth. \n" \
+			+ "You get a credits payment based on the amout of children you provide."
```
Firiman, line 13: 5 credits per birth + 2 per child up to a total limit of 30-40 credits per birth or so.

Skills/Perk/PregnancyMotherOfTheYear.gd
```diff
@@ -0,0 +1,24 @@
...
+func _init():
+	id = Perk.PregnancyMotherOfTheYear
+	skillGroup = Skill.Pregnancy
+
+func getVisibleName():
+	return "Mother of the year"
+
+func getVisibleDescription():
+	return "AlphaCorp can't thank you enough for providing more cannon fodder for their global operations. Oh! we meant to say what a nice kids you have we always happy to help them grow and prosper, right! \n" \
+			+ "You get a stats boost buff for 2 days after giving birth. \n" \
```
Firiman, line 12: For now this buff gives: +30 max pain + 30 max stamina + 30 max lust

Skills/Perk/PregnancyBroodmother.gd
```diff
@@ -0,0 +1,37 @@
...
+			+ "Increased cross species compatibility by 20% \n" \
+			+ "Fertility reduction medication has significantly less effect on you." #by 70%
+	
+func getSkillTier():
+	return 1
+
+func getPicture():
+	return "res://Images/Perks/Skill_Icon_PregnancyBroodmother.png"
+	
+#func isVisualToggleable():
```
Alexofp, line 25: I think a better idea would be to add ability to toggle any perk off or on after you 'learn' them. Perk that's toggled off would work as if its not even learned. Simple to implement and would work for every perk.

Yes, that way you will have to 'endure' the visual changes too if you want the bonuses. But I think it would make more sense. It would also allow us to make perks that have negatives too, people can just toggle them off

Hope I make sense. I can implement the toggling part

Skills/Perk/PregnancyBetterOvulationV2.gd
Alexofp: 450 sounds like a lot. Kids don't get processed but they do still take up some memory each. Not that much but still. Maybe lower it to 200 with both perks?..

Skills/Perk/PregnancyDesireToBreed.gd
```diff
@@ -0,0 +1,27 @@
...
+	return 0
+
+func getPicture():
+	return "res://Images/Perks/Skill_Icon_PregnancyDesireToBreed.png"
+
+#func getBuffs():
+#	if(!npc.menstrualCycle.isVisiblyPregnant()):
+#		return [
+#			buff(Buff.LustDamageBuff, [30]),
+#			buff(Buff.ReceivedLustDamageBuff, [20])
```
Alexofp, line 25: Ability to toggle perks would be useful here for example.

I think this perk is a bit unbalanced. Its basically an upgrade for characters that can't get pregnant. Although, I do understand that you gotta give births to earn experience for this skill? Still, maybe in the future you will be able to transform your pussy away or something x3

Maybe make it provide buffs only if in heat? Having heat would mean you can get pregnant

Skills/Perk/PregnancyMotherOfTheYear.gd
Alexofp: 5 is too much x3. Basically the way I see it, 1 work credit = 1 work day. And breeding is not that hard, make it 1 credit or something x3.

I'm not sure if giving credits is a good idea for a perk. The stat boost would be enough

Skills/Perk/PregnancyPainTolerance.gd
```diff
@@ -0,0 +1,25 @@
...
+	id = Perk.PregnancyPainTolerance
+	skillGroup = Skill.Pregnancy
+
+func getVisibleName():
+	return "Pain tolerance"
+
+func getVisibleDescription():
+	return "Life in prison is tough especially if you are constantly being pregnant, all the challenges that you have overcome made you much sturdier. \n" \
+			+ "Resistance to physical damage +15% \n" \
+			+ "Maximum stamina +30" #maybe add some percentage total, like +10% total stamina
```
Alexofp, line 13: Yeah I dunno. All the perks I make provide bonuses only under certain conditions. Nothing should be free

Skills/Perk/PregnancyProudMom.gd
```diff
@@ -0,0 +1,29 @@
...
+	id = Perk.PregnancyProudMom
+	skillGroup = Skill.Pregnancy
+
+func getVisibleName():
+	return "Proud mom"
+
+func getVisibleDescription():
+	return "You are looked from the better side by the prison staff. \n" \
+			+ "Get a free apple from the Nurse once per day. \n" \
+			+ "While pregnant gain some lust protection and have a lesser chance of the prison personnel to confront you." #base chance is 30% + exposure, this sets it to 10% + exposure
```
Alexofp, line 13: The free apple idea is so silly that I like it x3. I think the lust armor is not needed, just keep the apple

Skills/Perk/PregnancyShakyLegs.gd
```diff
@@ -0,0 +1,22 @@
...
+func _init():
+	id = Perk.PregnancyShakyLegs
+	skillGroup = Skill.Pregnancy
+
+func getVisibleName():
+	return "Shaky legs"
+
+func getVisibleDescription():
+	# For now this works as - if you see a 20+ cm dick you go in heat, might make it chance based depending on the length
+	return "You are wild about big hard juicy cocks and the thoughts of getting filled and knocked up by them. Seeing one, your legs become all shaky and you are forced into heat."
```
Alexofp, line 12: I.. I dunno. The implementation would be a hacky nightmare and it's kinda unrealistic (even more than usually x3). A simplier idea would be something like: Receive heat if you orgasm from someone with a big dick. Would need some code changes too but way less I think

But I suggest rethinking this one..

Skills/Skill/Pregnancy.gd
```diff
@@ -0,0 +1,17 @@
+extends SkillBase
+
+func _init():
+	id = Skill.Pregnancy
+
+func getVisibleName():
+	return "Pregnancy"
```
Alexofp, line 7: How about "Reproduction"

Firiman: Procreation or reproduction :thinking:

Skills/Perk/PregnancyMotherOfTheYear.gd
Firiman: 1 work day = 1 credit is interesting, why does empty plastic bottle cost 5 credits then? xD Also prostitutes charge like 5-8 credits up to 20+ in some cases. So there should be either a fractions of a credit used and prices lowered significantly for the most items or increase it to something like 5 credits per workday.

Firiman: Default preg time is 5 days, so 1 is kinda low for tier 3 perk which you can get only by giving birth plenty of times. Maybe 3 + 1 per child with a limit of 20?

Skills/Perk/PregnancyBroodmother.gd
Firiman: I still think toggling only visual is the way. Remember that the perks can have requirements of the other perks. So we would need to toggle off the whole chain in some cases. Also imagine situation where perk gives mostly negative bonuses but is a requirement for a better perk in the next tier, just toggling it off without toggling off better perk would give an advantage for no reason. Also toggling perks on/off might be a source for some bugs with certain perks or exploits in the future.

Skills/Perk/PregnancyPainTolerance.gd
Firiman: > Yeah I dunno. All the perks I make provide bonuses only under certain conditions. Nothing should be free

Make giving birth to like 50-100 children as an additional requirement?

Skills/Perk/PregnancyMotherOfTheYear.gd
Alexofp: Everyone is greedy and AlphaCorp really doesn't want to give inmates credits ^^. Thats why inmates/vendomats can ask for a lot while the corporation gives very little. I dunno, I would still just keep the stats boost and remove the credits reward

You know what would be cool? A 'bounty' board in the nursery that generates random tasks like "Give birth/be a father to 3 cats" or "give birth to 5 males" or whatever. Tasks that ask for hybrids could pay more. And basically, even if you're not the right species, you can still make that happen with strapons and cum transfering x3. Could also have tasks like "donate cum/breastmilk"

So basically I think giving credits for kids is too easy, needs more challenge.. But I like the temporary stats boost ^^

Skills/Perk/PregnancyBroodmother.gd
Alexofp: I still like my idea x3. Yeah, perks that depend on each other is an issue. But it's fairly easy to solve, you either dissallow that or make it so all the 'child' perks get toggled off automatically.

Basically I'm not a huge fan of visual toggle but I can stomach it if we also add a full toggle x3

Skills/Skill/Pregnancy.gd
Alexofp: waaait. How about Fertility skill. Fertility skill sounds perfect I think, even though it's a name of a buff already I think, who cares x3

Skills/Perk/PregnancyMotherOfTheYear.gd
Firiman: That nursery board is a nice idea indeed. Although it should be unrelated to perks IMHO. 
There is a difference still between greed and a comical level of greed where an empty bottle costs 5 credits but you get paid 1 for a work day. Also I had the impression that AlphaCorp is really interested in inmates producing more children for their war efforts or whatever, that's why I was wondering why condoms and infertility pills are legal. Maybe make some shady character to sell all this stuff like condoms, anaphrodisiac pills, birth control pills and if you are caught by the guards with them you get fed breeder pill and put into stocks.

Skills/Perk/PregnancyBroodmother.gd
Firiman: > I still like my idea x3. Yeah, perks that depend on each other is an issue. But it's fairly easy to solve, you either dissallow that or make it so all the 'child' perks get toggled off automatically.

Ok, automatic toggling off dependent perks is the way then, but perks should be distinguishable between toggled off and not learnt. Visual effects toggle is not needed in that case.

Firiman:
Current implementation as in the mod, not ready for review yet.

Alexofp:
A person in discord gave an interesting idea for a perk

Bassically a perk that allows you to ovulate even if already pregnant. Would allow you to get even more pregnant x3

Not too hard to implement too, would just need a perk check in https://github.com/Alexofp/BDCC/blob/611a47575ef023ca1ca65d550ca88253a979181a/Game/Pregnancy/MenstrualCycle.gd#LL65C61-L65C61

so it would look like
`if(!ovulatedThisCycle && cycleProgress >= willOvulateAt && (!isPregnant() || getCharacter().hasPerk(Perk.ASdasd))):`

might as well let them experience heat too while pregnant, just needs a simillar change for the `func isInHeat():` function

Firiman:
Current implementation as in the mod, not ready for review yet.

> A person in discord gave an interesting idea for a perk
> 
> Bassically a perk that allows you to ovulate even if already pregnant. Would allow you to get even more pregnant x3
> 
> Not too hard to implement too, would just need a perk check in https://github.com/Alexofp/BDCC/blob/611a47575ef023ca1ca65d550ca88253a979181a/Game/Pregnancy/MenstrualCycle.gd#LL65C61-L65C61
> 
> so it would look like `if(!ovulatedThisCycle && cycleProgress >= willOvulateAt && (!isPregnant() || getCharacter().hasPerk(Perk.ASdasd))):`
> 
> might as well let them experience heat too while pregnant, just needs a simillar change for the `func isInHeat():` function

I've seen that, but how feti is going to grow inside the belly if you just keep going? Sounds like you could just prolong pregnancy indefinitely with this feature.  For example say you got pregnant, feti from your first ovulation grew to 60% towards birth, then you ovulated and got pregnant again, when the 1st batch reaches 100% readiness for birth, your second batch is only at 40%. Now if we wait for the second batch to grow to 100% the first one would be at 160% and if we repeat that multiple times, the very first batch might reach 1000% or more and that's kinda silly.
 Now we may add some conditions like if there is a cell that grew to certain percent towards birth, no new ovulation is possible. But that wouldn't change things much from how the system currently works.

Alexofp:
that part isn't realistic already, feti stop growing after 100%, allowing you to give birth at any time you want. I dunno, I think it would be fun and doesn't require much coding ^^

Firiman:
> that part isn't realistic already, feti stop growing after 100%, allowing you to give birth at any time you want. I dunno, I think it would be fun and doesn't require much coding ^^

Of course we aren't talking about realism here. Im interested to know what's going to happen if we have some feti that are not ready for birth, but we give birth to the ones who are ready? Are those who aren't ready die or you just keep your pregnancy?

Alexofp:
with the current code they would get born anyway like they're ready x3 which is not the worst solution I think. I guess the giving birth scene could be modified a bit so it doesn't allow you to give birth until every fetus is ready

Skills/Perk/PregnancyMotherOfTheYear.gd
Firiman: Lowered to 2 credits per birth + 1 per child (limit is 20) for now.

Skills/Perk/PregnancyPainTolerance.gd
Firiman: Actually there are quite a few "free" bonuses in combat and some other perk trees. Lowered physical resistance to 10% for now. It's not really that much for a T3 perk. BDSM perk is giving 15% but is T1 although with a condition. So I think this is balanced. But adding some req like 50-100 children born is possible of course.

Skills/Perk/PregnancyBetterOvulationV2.gd
Firiman: Lowered to 250 with both perks (100 base + 50 first + 100 second) for now.

Firiman:
7/9 perks ready
FertilityBetterOvulationV3 (Anomalous pregnancy) and FertilitySubmissiveAndBreedable left

Skills/Perk/PregnancyShakyLegs.gd
Firiman: Replaced with Submissive and breedable perk.

Firiman:
Full review is not necessary atm just check if Submissive And Breedable perk is okay before I start implementing it.

Game/Pregnancy/Child.gd
```diff
@@ -9,9 +10,23 @@ var fatherID: String
...
+const EGG_MONOZYGOTIC_LEVEL = {1: "", 2: "Twins", 3: "Triplets", 4: "Quadruplets", 5: "Quintuplets", 6: "Sextuplets"}
 
 func generateUniqueID():
 	uniqueID = GlobalRegistry.generateChildUniqueID()
+	
+func setBornFromMonozygoticStatus(monozygotic):
+	bornFromMonozygotic = EGG_MONOZYGOTIC_LEVEL[monozygotic]
+	
+func generateName():
+	match gender: "male", "peachboy":
```
Alexofp, line 23: better to use `NpcGender.Male`, `NpcGender.Peachboy`. Or you can add a new static function to NpcGender like `isMasculine` that will return true for those two genders

Game/Pregnancy/MenstrualCycle.gd
```diff
@@ -113,8 +125,14 @@ func processTime(seconds):
...
 			ovulate()
 	
+	var readyFetusAmount: = 0
 	for egg in impregnatedEggCells:
 		egg.processTime(seconds)
+		if(egg.fetusIsReadyForBirth()):
+			readyFetusAmount += 1
+	
+	if(readyFetusAmount == impregnatedEggCells.size() && impregnatedEggCells.size() > 0):
+		onAllFetusReadyForBirth()
```
Alexofp, line 135: this makes it so this function is called on every process time call. Before it was called when one of the eggcells was ready. This could be kinda slow.

Inventory/Items/AppleItem.gd
```diff
@@ -6,6 +6,9 @@ func _init():
 
 # apple is illegal only temporally as a test item
 func getVisibleName():
+	if(GM.pc != null):
+		if(GM.pc.hasPerk(Perk.FertilityProudMom) && getWearer() == GM.pc):
+			return "Tasty Apple"
```
Alexofp, line 11: I'd rather just create a separate apple item honestly x3. Because right now this perk 'legalizes' apple stealing x3. Or you can add a variable 'notStolen' and set it to true for the perk apples.

Inventory/Items/UsedCondom.gd
```diff
@@ -26,6 +26,11 @@ func getCondomBreakChance():
 	if(!OPTIONS.isContentEnabled(ContentType.RiskyCondoms)):
 		return 0.0
 	
+	# Check to make sure if PC is fucked with his own condoms still worked with the perk
+	var wearerIsNotPC: bool = getWearer() != GM.pc || (!isWornByWearer() && getWearer() == GM.pc)
+	if(wearerIsNotPC && GM.pc.hasPerk(Perk.PregnancyDesireToBreed)):
```
Alexofp, line 31: forgot to rename `Perk.PregnancyDesireToBreed` to `Perk.FertilityDesireToBreed`

Inventory/Items/Condom.gd
```diff
@@ -32,6 +32,11 @@ func getCondomBreakChance():
 	if(!OPTIONS.isContentEnabled(ContentType.RiskyCondoms)):
 		return 0.0
 	
+	# Check to make sure if PC is fucked with his own condoms still worked with the perk
+	var wearerIsNotPC: bool = getWearer() != GM.pc || (!isWornByWearer() && getWearer() == GM.pc)
+	if(wearerIsNotPC && GM.pc.hasPerk(Perk.FertilityDesireToBreed)):
+		return 90.0
```
Alexofp, line 38: I'm not sure about this. What if I add a scene in the future where an npc fucks some other npc with a condom.. This would need to be handled differently then

Modules/MedicalModule/Nursery/NurseryTalkScene.gd
```diff
@@ -39,8 +39,17 @@ func _run():
 		else:
 			saynn("You approach the nurse and try to get her attention. She drags her gaze away from the screen and looks at you.")
 
+			# (if player has the perk FertilityProudMom and didn't get an apple today)
+			if(GM.pc.hasPerk(Perk.FertilityProudMom) && !GM.main.getModuleFlag("MedicalModule", "Nursery_hadFreeAppleFromPerkToday", false)):
+				saynn("[say=nurse]Oh it's you again! I try to make sure that every mom here stays on the healthy side, take this[/say]")
+				saynn("She gives you an apple")
+				
+				GM.pc.getInventory().addItem(GlobalRegistry.createItem("appleitem"))
```
Alexofp, line 47: this could be something like
```
var apple = GlobalRegistry.createItem("appleitem")
apple.setNotStolen(true)
GM.pc.getInventory().addItem(apple)
```

Alexofp: Oh, btw, its a bad idea to do this in the _run function. Ideally the run function shouldn't change the state of the game ever. I'm a little guilty of that but I try to keep to only flag changes. Giving an apple is changing the state, even if it shouldn't break anything. How to solve that? Add a `_reactInit` function that gives the apple and also sets some variable like `gaveApple`. Then you just check that variable here in the run function

Player/Player3D/Doll3D.gd
```diff
@@ -319,7 +319,10 @@ func setBreastsScale(breastsScale: float):
 	setBoneScaleAndOffset("DeformBreasts", breastsScale, Vector3(0.18713, 0.199727, 0.0)*mul)
 
 func setPregnancy(progress: float):
-	setBoneOffset("DeformBelly", Vector3(-0.03244, 0.706324, 0.0)*progress)
+	if(GM.pc.hasPerk(Perk.FertilityBroodmother)):
```
Alexofp, line 322: This will deform bellies for every pregnant character if the player has the perk x3

instead of this check, you gotta add a function argument like `hasBigBelly: bool = false` and then you pass a true into this function from inside the BaseCharacter's softDollUpdate function if that character has the perk

Scenes/SceneBase.gd
```diff
@@ -50,6 +51,9 @@ func initScene(args = []):
 	_initScene(args)
 	_reactInit()
 	
+	if(args.has("pc")):
+		sceneHasPC = true
```
Alexofp, line 55: nah, that's not how you do it. Your best would be to make the `isSexSceneWithPC` return false by default and then override that behavior for the procedural sex scene and other scenes that you need

StatusEffect/Pregnant.gd
```diff
@@ -38,7 +38,10 @@ func getIconColor():
 
 func getBuffs():
 	var progress = character.getPregnancyProgress()
-	
+	var modifier = 1
+	if(GM.pc.hasPerk(Perk.FertilityBroodmother)):
```
Alexofp, line 42: `if(character.hasPerk(Perk.FertilityBroodmother)):`

Alexofp (commented):
meow meow meow. My inner optimization gremlin cries from the amount of babies that people will have x3. It's like I should have made them just be a number

Game/Pregnancy/MenstrualCycle.gd
```diff
@@ -299,12 +328,15 @@ func giveBirth():
...
 		var eggCell: EggCell = egg
 		
-		var newChild: Child = Child.new()
-		newChild.generateUniqueID()
-		newChild.loadFromEggCell(eggCell)
-		newChild.setBirthday(GM.main.getDays())
+		for x in eggCell.monozygotic:
+			var newChild: Child = Child.new()
+			newChild.generateUniqueID()
+			newChild.loadFromEggCell(eggCell)
```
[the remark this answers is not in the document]
Firiman: Right, I can change that if needed.

Inventory/Items/AppleItem.gd
Firiman: You still can be assaulted by the guards in the process of stealing like normal. I was thinking about making a separate "legal" apples, but that made no sense to me because stolen or not apples look the same. If you are still concerned I can make Nurse give the player apple juice instead.

Firiman:
> meow meow meow. My inner optimization gremlin cries from the amount of babies that people will have x3. It's like I should have made them just be a number

You can just set egg cell lifespan very high and and short cycle length in the base game and have 100 babies or even more per birth. Actually I was thinking about tweaking the code so any non impregnated cells get deleted when the cycle ends to prevent this madness.

Game/Pregnancy/MenstrualCycle.gd
Firiman: I checked readyToGiveBirth signal is never used after this commit 
https://github.com/Alexofp/BDCC/commit/49602bb2e9056017593462cb35af3daabadce38b
So it is only a function call which do nothing after the first time it is called.

Game/Pregnancy/EggCell.gd
```diff
@@ -168,3 +187,5 @@ func loadData(data):
 	motherSpecies = SAVE.loadVar(data, "motherSpecies", ["feline"])
 	resultSpecies = SAVE.loadVar(data, "resultSpecies", [])
 	resultGender = SAVE.loadVar(data, "resultGender", NpcGender.Male)
+	monozygotic = SAVE.loadVar(data, "monozygotic", 1)
+	_fetusReadyForBirth = SAVE.loadVar(data, "fetusReadyForBirth", false)
```
Alexofp, line 191: Please don't begin the variables with _ in the future. Looks bad when some use it and some don't like here

Game/Pregnancy/MenstrualCycle.gd
```diff
@@ -54,15 +54,24 @@ func getCurrentStage():
 		return CycleStage.Ovulation
 	return CycleStage.LutealPhase
 
+func isEligibleForProlongedPregnancy() -> bool:
+	if(!getCharacter().hasPerk(Perk.FertilityBetterOvulationV3)):
+		return false
+
+	for egg in impregnatedEggCells: #to prevent perpetual pregnancy. If any egg reaches 2.5 progress
+		if(egg.getProgress() >= 2.5): # no further ovulations allowed
```
Alexofp, line 62: disabling this limit could be another perk, you know :p

Game/Pregnancy/MenstrualCycle.gd
```diff
@@ -77,7 +86,11 @@ func getCharacter():
 func initCycle():
 	cycleProgress = RNG.randf_range(0.0, 1.0)
 	
-func newCycle():
+func newCycle(shouldClearNonPregEggs: = true):
+#	print(getCharacter().getName(), " Entered new cycle" )
+	if(shouldClearNonPregEggs):
+		for orificeType in OrificeType.getAll():
+			eggCells[orificeType] = []
```
Alexofp, line 93: This is kinda unrealistic, why not keep the old behaviour and instead make it so you can't get into heat if you ovulated this cycle already. This would fix the problem with too many eggcells

Game/Pregnancy/MenstrualCycle.gd
```diff
@@ -134,11 +157,16 @@ func ovulate():
...
-			
+		
+		var ch = getCharacter()
 		var amountOfEggs = RNG.pickWeightedPairs(possibleEggAmounts)
-		if(amountOfEggs == null):
-			amountOfEggs = 1
-		#print("AMOUNT OF EGGS: "+str(amountOfEggs))
+		amountOfEggs = max(ch.getMinEggsAmount(), int(round(amountOfEggs * ch.getEggsBonusMod())))
+		if(GM.pc.hasPerk(Perk.FertilityBetterOvulation) && amountOfEggs < 10):
+			amountOfEggs += RNG.randi_range(0, 4) #otherwise species with low base eggs like humans, won't get much bonus
```
Alexofp, line 165: I would just use ceil() instead of round() for the amountOfEggs variable instead of hardcoding it like that

Game/Pregnancy/MenstrualCycle.gd
```diff
@@ -190,18 +218,26 @@ func getPregnancyProgress() -> float:
...
-func isReadyToGiveBirth():
-	return getPregnancyProgress() >= 1.0
+func isReadyToGiveBirth() -> bool:
+	var readyFetusAmount: = 0
+	for egg in impregnatedEggCells:
+		if(egg.fetusIsReadyForBirth()):
+			readyFetusAmount += 1
+			
+	if(impregnatedEggCells.size() > 0):
+		return readyFetusAmount == impregnatedEggCells.size()
```
Alexofp, line 228: I usually expand it into the full if(a == b) return true else return false. Easier for me to read.
I also suggest replacing == with >=. Yeah, it won't change anything but it feels better for me
Also just a note but this means you can technically go from 'being ready to give birth' back to not being ready if one of the eggcells got impregnated late. Just wondering if that's intended

Game/Pregnancy/MenstrualCycle.gd
```diff
@@ -276,12 +311,11 @@ func getRoughChanceOfBecomingPregnant() -> float:
...
 	return roughChance * 100.0
 
-func onEggCellReadyForBirth(_egg):
-	#print("EGG READY TO BIRTH")
-	emit_signal("readyToGiveBirth")
-	if(!noticedReadyToGiveBirth):
-		noticedReadyToGiveBirth = true
-		emit_signal("readyToGiveBirthOnce")
+#func onAllFetusReadyForBirth():
+#	emit_signal("readyToGiveBirth") #NOT USED AFTER https://github.com/Alexofp/BDCC/commit/49602bb2e9056017593462cb35af3daabadce38b
```
Alexofp, line 315: well if its dead code its better to remove it fully, no need to leave the commented code

Util/RNG.gd
```diff
@@ -149,3 +149,19 @@ static func randomFemaleName():
...
+	var character
+	if(characterPenetrated is String):
+		character = GM.main.getCharacter(characterPenetrated)
+	else:
+		character = characterPenetrated
+	
+	if(character.hasPerk(Perk.FertilityDesireToBreed)):
+		return chance(clamp((chance * 4.5), 1, 95)) #limits max chance to break to 95%
+	else:
+		return chance(chance)
```
Firiman, line 167: Could be moved somewhere else to Util for example, if you want.

Game/SexEngine/FetishHolder.gd
```diff
@@ -33,6 +33,15 @@ func setCharacter(newchar):
 func setFetish(fetishID, interest):
 	fetishMap[fetishID] = interest
 
+func modifyFetishByTier(fetishID, numTiers: int = 1) -> bool:
```
Alexofp, line 36: I'd call it incrementFetish(). Couldn't understand what the function does just by the name, had to read the code

Game/SexEngine/FetishHolder.gd
```diff
@@ -33,6 +33,15 @@ func setCharacter(newchar):
...
+	var currentInterestValueNum = FetishInterest.interestToNumber(getFetishInterest(fetishID))
+	var projectedInterestValueNum = int(clamp((currentInterestValueNum + numTiers), -4, 4))
```
Alexofp, line 38: I don't like the -4 and 4. Too hardcoded, might break. Best way to solve that is to add a clampInterestValue or something to FetishInterest class. It would still be hardcoded but at least its all would be in one file

Game/SexEngine/SexActivity/SexVaginalOnAllFours.gd
```diff
@@ -808,6 +816,13 @@ func getSubOrgasmHandlePriority():
 		return 10
 	return 5
 
+func fetishesToMod():
```
Alexofp, line 819: This is too much for this pull request. Please remove the fetish modifying, way outside of the scope of a pull request that adds a new skill

Game/SexEngine/SexEngine.gd
```diff
@@ -281,14 +283,21 @@ func generateGoals():
 					
 					if(sexGoal.isPossible(self, personDomInfo, personSubInfo, goalData) && !sexGoal.isCompleted(self, personDomInfo, personSubInfo, goalData)):
 						possibleGoals.append([[goal[0], sub.getID(), goalData], goal[1]])
-		
+						if((goal[0] == SexGoal.FuckAnal || goal[0] == SexGoal.FuckVaginal) && submissiveAndBreedablePerkHelper):
```
Alexofp, line 286: Not a fan of it being hardcoded like that. Dunno how would I do it though

Skills/Perk/FertilityDesireToBreed.gd
```diff
@@ -0,0 +1,27 @@
...
+	id = Perk.FertilityDesireToBreed
+	skillGroup = Skill.Fertility
+
+func getVisibleName():
+	return "Desire to breed"
+
+func getVisibleDescription():
+	return "Your desire to be bred is so high that new life finds its way inside you with ease. \n" \
+			+ "While not pregnant you deal +30% lust damage and become more susceptible to the lust yourself. \n" \
+			+ "Significantly increases the chances of condoms breaking inside you." #450% to break results in 90% for normal and 4.5% for quality condoms
```
Alexofp, line 13: I would remove the lust buffs/debuffs and just keep the condoms breaking. Could reword the description so its about the player purposfully sabotaging condoms because right now its kinda magical and doesn't explain anything

Skills/Perk/FertilitySubmissiveAndBreedable.gd
```diff
@@ -0,0 +1,22 @@
...
+func _init():
+	id = Perk.FertilitySubmissiveAndBreedable
+	skillGroup = Skill.Fertility
+
+func getVisibleName():
+	return "Submissive and breedable"
+
+func getVisibleDescription():
+	return "Others view you as a subservient, perfect toy to please their carnal desires, most noticeably breeding. " \
+			+ "Dominant partners may go an extra round or two, and gain a breeder fetish if not having one already."
```
Alexofp, line 12: remove the fetish gaining. I have others plans for how to implement that, just doing it off of a perk is too easy and unrealistic

Util/RNG.gd
```diff
@@ -149,3 +149,19 @@ static func randomFemaleName():
 static func pickHashed(ar, thehash:int):
 	thehash = thehash % ar.size()
 	return ar[thehash]
+
+static func shouldCondomBreakWhenCumInside(characterPenetrated, chance: float = 20.0) -> bool:
```
Alexofp, line 153: Move this function into the baseCharacter class. RNG.gd is just a utlity class that is not supposed to know -anything- about the game. Basically you should be able to grab this file and add it into a new project and it should just work. Right now it won't

Game/Pregnancy/Child.gd
```diff
@@ -79,6 +99,8 @@ func saveData():
 	return data
 
 func loadData(data):
+	name = SAVE.loadVar(data, "name", "")
```
Alexofp, line 102: if the loaded name is empty you should generate a new random one
name = SAVE.loadVar...
if(name == ""):
    generateName()

Alexofp (commented):
I understand that it's hard but we should try to make the code cleaner with each change. Otherwise there will be bugs and crashes and it will be hard to expand the game. Sorry if I sound grumpy.

Inventory/Items/AppleItem.gd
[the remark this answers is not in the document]
Firiman: I was thinking about making a separate "legal" apples, but that made no sense to me because stolen or not apples look the same. But I can make Nurse give the player apple juice instead.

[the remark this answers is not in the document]
Alexofp: if isNotStolen:
desc += "\nThis particular one has AlphaCorp's sticker of approval"

could be something like that.

Game/SexEngine/FetishHolder.gd
Firiman: You mean the current system from "Hates" to "Loves" with 9 states could change?

Alexofp: avoiding hardcoded numbers is clean code. The answer is yes, it could change at any moment

Game/Pregnancy/MenstrualCycle.gd
Firiman: It is intended yeah.

Game/Pregnancy/MenstrualCycle.gd
Firiman: Madness xd

Game/Pregnancy/MenstrualCycle.gd
Firiman: Unrealistic? But not fertilised eggs are shed during menstruation phase.

Skills/Perk/FertilitySubmissiveAndBreedable.gd
Firiman: I asked a week ago if Submissive and Breedable perk is okay, but you didn't reply. https://github.com/Alexofp/BDCC/pull/14#issuecomment-1567950657

Alexofp: I was busy/tired and forgot, sorry

Alexofp:
Thank you <3

PR #16, opened 2023-06-20
Uhhh- idk how to title these TTwTT

NielIsOnGithub:
Updated StatusEffect.InHeat to encompass "In Rut" (for those with penises) and "In Heat and In Rut" (for those with penises and vagina). Basically just added text.

Added StatusEffect.EverLeaking, has a 1% chance every 10 second to give the player StatusEffect.InHeat. This effect is given to the player when their current lust is greater than or equal to 85% of their max lust.


StatusEffect/InHeat.gd
```diff
@@ -5,7 +5,18 @@ func _init():
...
+		character.hasBodypart(BodypartSlot.Vagina)
+	]
+	
+	match toCheck:
+		[false, true]:
+			return "In Heat"
+		[true, false]:
+			return "In Rut"
+		[true, true]:
+			return "In Heat and In Rut"
```
Alexofp, line 19: This is a strange way to check that, I would prefer just lots of if-elif checks.

But if you want rut its better to make it its own status effect. Right now its just renamed heat and works exactly like it (that means only girls have it)

StatusEffect/EverLeaking
```diff
@@ -0,0 +1,57 @@
...
+
+func getIconColor():
+	# light pink?
+	return IconColorDarkPurple
+
+func getBuffs():
+	return [
+		buff(Buff.ReceivedLustDamageBuff, [15]),
+		buff(Buff.AmbientLustBuff, [30]),
+		buff(Buff.ExposureBuff, [30]),
```
Alexofp, line 56: Right now this status effect is a punishment for having high lust. It's unavoidable (heat is also unavoidable yeah. But heat has some functionality like showing you when your most unsafe days are)

StatusEffect/EverLeaking
```diff
@@ -0,0 +1,57 @@
...
+		
+		turns += 10
+
+func getEffectName():
+	return "Ever-Leaking"
+
+func getEffectDesc():
+	var message = "You're too horny. "
+	match _toCheck:
+		[true, false, false]: # has penis
```
Alexofp, line 31: I'd just prefer if checks.. You even realize that [true, false, false] isn't descriptive at all so you added comments that explain them. Why not just use the actual functions instead of random arrays of boolean flags? x3

StatusEffect/EverLeaking
```diff
@@ -0,0 +1,57 @@
+extends StatusEffectBase
+
+var _toCheck = [
+	character.hasBodypart(BodypartSlot.Penis),
+	character.hasBodypart(BodypartSlot.Vagina),
+	character.hasWombIn(BodypartSlot.Anus)
```
Alexofp, line 6: I'm surprised that this works (if it does)
also you forgot to add .gd at the end of the file

Alexofp (commented):
Thank you for choosing to code something simple!

Sadly I don't think this status effect fits the game, it will just bloat the status effect panel without adding any interesting gameplay. Basically it's punishing you for having high lust by making it so you take even more lust damage x3. I hope you understand my reasoning ^^

Also, please just use `if(character.hasPenis()):` instead of those scary match statements next time, makes the code easier to read

PR #17, opened 2023-06-26
Update ParadedOnALeashScene.gd

NielIsOnGithub:
Added a new parameter called "teleportwhenskipped" and a few lines of code that made it so that the player is teleported to the "endLocation" given when they press the skip button.


Alexofp:
Looks alright ^^. But you can make so they're teleported to the end by default

Alexofp:
Thank you ^^

PR #25, opened 2023-09-01
Update CurveRenderer.gd

PeteTimesSix:
* Cleanup
 * Allow specifying origin (defaults to self if left empty) for placing chains into scenes that connect two arbitrary points
 * Remove gaps between segments (by averaging neighboring segments together)
 * A few comments


Alexofp:
Looks good ^^ I simplified the API a lot, dunno if you saw x3. Still have to go through each scene that mentions a leash and add it

PR #26, opened 2023-09-19
Add buffs and skills into module resolution

PeteTimesSix:
Very simple, but seems that It Just Works (see the first commit for a quick test).

Dunno why its not already there, I imagine its an oversight? In any case you fail to commit every pull request you don't make, so...


Alexofp:
Yeah I just got lazy with it, heh. Thanks

PR #27, opened 2023-09-19
Scriptable skills/perks

PeteTimesSix:
* Support for skills that do not (visibly) use exp (for skill trees with perks granted by scripts/scenes/etc)
 * Support for perks that cannot be toggled by the player
 * Support for perks that cannot be unlocked by the player
 * Support for perks that are invisible when locked
 * Support for perks that are invisible when unlocked

 * All of the above used to make the starting perks visible
 
 This one's a little more of an additional feature than filling a preexisting gap, but I think it should be useful. I figured I might as well scrape it off the mod I've been experimenting with and PR it.


Alexofp:
> All of the above used to make the starting perks visible

is that really that useful though. For me it feels like it just clutters the skills tab

I'm not sure there will even be skills/perks granted by the scenes x3. I don't have ideas for any. I think fallout 4 had perks that you unlock by going through some quests. I don't think it worked very well x3. You're free to try to change my mind x3

Player/Player.gd
```diff
@@ -62,6 +62,8 @@ func _ready():
 	#giveBodypart(GlobalRegistry.createBodypart("caninepenis"))
 	updateNonBattleEffects()
 	
+	skillsHolder.addSkillExperience(Skill.Start, 0)
```
Alexofp, line 65: This is very hack-ish. What if `addSkillExperience` changes so it doesn't add the skill if the exp is 0. I'd prefer some kind of `alwaysVisible()` func or something for the skill then. But I'm still not sure if showing a skill that doesn't have anything unlockable is a good idea. Player might assume that all skills will be like that

PeteTimesSix:
Re: hackish - yes it is. Tends to be the case with whichever line of code I write last - those dont go through the usual second/third guessing process :p

I think its not particularily user-friendly that its currently not possible to review which starting perks a character has either (what if someone loads a six months old save and cant remember if they made that character infertile, for example?). Honestly, the starting perks are a bit of a special case. Maybe they should be in a separate tab, or separated by a horizontal line? Or I can just take it out entirely if you dont like the idea, its your game of course :D

I personally find gaining perks by interacting with the game world (kind of like, yes, Fallout 4's skill magazines, but another example that comes up often is being taught special moves by martial art masters on top of mountains) more interesting than picking from an abstract list. To be fair though BDCC is halfway there already with its skill-specific exp. 

The idea that started me down this path is a hypnosis skill tree which the player picks bonuses from as normal, but with every tier they unlock they also get a drawback perk (which I therefore dont want the player to unlock manually or toggle off).

Anyway, Ill poke at it a bit more once I get back from being the boring, corporate kind of drone and get back to you then.

PeteTimesSix:
Is this more to your liking?

Having Start as an actual skill is still a placeholder while I look into what it'd it take to implement a separate tab for it.

PeteTimesSix:
Okay, here's my final(ish) take on it. Feel free to pick out whichever bits (if any) of it you like, and I'll scrape out the rest.

Alexofp:
I will take a look when I'm less tired, thank you

Alexofp:
Thank you. I will probably do some tweaks after merging if you don't mind

PR #29, opened 2023-10-09
Add more stuff into module resolution

PeteTimesSix:
Added the following as classes modules can register:
- stageScenes
- lootTables
- lootLists
- fetishes
- sexGoals
- sexActivities
- sexTypes
- fluids

Guessing those should be fine.

Left out stats, skins, imagePacks, lustTopics (not sure if the game can handle extras with those), mapFloors (register function for those has two params) and fightClubFighters (not sure those are actually a separate thing from characters).

### DIFF
--- a/Modules/Module.gd
+++ b/Modules/Module.gd
@@ -19,6 +19,14 @@ var worldEdits = []
 var gameExtenders = []
 var computers = []
 var partSkins = []
+var stageScenes = []
+var lootTables = []
+var lootLists = []
+var fetishes = []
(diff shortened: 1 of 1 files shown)


PeteTimesSix:
...also, I should probably have asked first, but do you prefer PRs on the main branch or the dev branch?

Alexofp:
Doesn't matter too much. I can pull stuff between branches. But you should learn how to create branches rather than committing into your own main :p. Thanks though

PR #30, opened 2023-10-31
Attack-specific reaction animation

PeteTimesSix:
Allow for attacks to override the reaction animation. I'm using it for an attack that makes the enemy kneel, but I imagine it might be useful, for example tasers making a character twitch or flashbangs making them cover their eyes, that sort of thing.

### DIFF
--- a/Attacks/BaseAttack.gd
+++ b/Attacks/BaseAttack.gd
@@ -41,13 +41,14 @@ func doAttack(_attacker, _receiver, _context = {}):
 		result["text"] = "!No attack text provided!"
 	
 	result["attackerAnimation"] = getAttackSoloAnimation()
+	var receiverReactAnimation = getAttackHitReactAnimation(result)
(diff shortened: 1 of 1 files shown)


Attacks/BaseAttack.gd
```diff
@@ -295,6 +296,12 @@ func checkDodged(_attacker, _receiver, _damageType, customDodgeMult = 1, minChan
 
 func getAttackSoloAnimation():
 	return ""
+	
+func getAttackHitReactAnimation(result):
```
Alexofp, line 300: maybe you could pass the _receiver (and _attacker too, why not) into this function and then move all the if isBlocking: anim="blocking" checks into this function as well. Feels strange to only be able to override the hurt animation

PeteTimesSix: Fair point. Changes made.

PR #31, opened 2023-10-31
More granular stat change signals

PeteTimesSix:


Alexofp:
Any reason for these signals? I don't mind new code but it should have a reason ideally x3

PeteTimesSix:
My use case is I've got a status effect (being hypnotized) that I knock stacks off of when the character gets punched. Other examples... Well, a traditional one is charging some sort of rage meter when taking damage? You could have some perk that automatically uses an energy drink when your stamina drops too low?

Alexofp:
Perks/status effect don't really have access to signals, I do everything through just functions. I treat signal as something internal. Core game code can use them but high-level objects like perks shouldn't really.

PeteTimesSix:
That's a shame. Signals seem more modding-friendly from what I've seen so far, since you only need access to the instance to hook into them rather than having to rewrite the function itself to call modded code. That being said, I can work around it in this case anyway, so...

Game/BaseCharacter.gd
```diff
@@ -3,6 +3,9 @@ class_name BaseCharacter
 
 #Base class that has all the functions
 signal stat_changed
+signal pain_changed(amount)
+signal lust_changed(amount)
+signal stamina_changed(amount)
```
Alexofp, line 8: make these provide the old and new values rather than just a delta and I can merge ^^

PeteTimesSix: I was actually going to do that at first before I decided to be minimalist for some reason :D
Done.

Alexofp:
thankies

PR #35, opened 2023-11-21
items can react to forceful equip

PeteTimesSix:
Give items the ability to trigger effects when force-equipped. Pass through the attacker to check for perks and such.

### DIFF
--- a/Attacks/NpcAttacks/CuffPCHands.gd
+++ b/Attacks/NpcAttacks/CuffPCHands.gd
@@ -19,7 +19,7 @@ func _doAttack(_attacker, _receiver, _context = {}):
 	if(checkDodged(_attacker, _receiver, DamageType.Physical)):
 		return genericDodgeMessage(_attacker, _receiver)
 	
-	_receiver.getInventory().forceEquipStoreOtherUnlessRestraint(GlobalRegistry.createItem("inmatewristcuffs"), _attacker)
+	_receiver.getInventory().forceEquipByStoreOtherUnlessRestraint(GlobalRegistry.createItem("inmatewristcuffs"), _attacker)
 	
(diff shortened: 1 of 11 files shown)


PeteTimesSix:
Yeah, there are - but I think I covered all the dynamic ones, and my thinking is that in scenes it can/should? be handled by the scene.

Alexofp:
Depends on what you need it for

PeteTimesSix:
Could go over them and pass the relevant pc/npc through if you think that's preferable though, it wouldnt take that long - but yeah, that's my thinking, in scenes you're usually equipping a specific item, so you know if it overrides afterForceEquip and can pass the npc through if necessary

in case you're wondering what my use case is, its adding a base level of hypnosis when forcing visors onto people with certain perks: [func on_forceEquip(_attacker):](https://github.com/PeteTimesSix/BDCC/blob/de70c12928685a316ba27710c321873f43cd85ef/Modules/HypnokinkModule/Items/HypnoVisor/HK_Visor.gd#L95)

Alexofp:
Maybe it's better to have a `onEquippedBy(otherCharacter, forced=false)` function on the item then. Just for the future. Don't worry about modifying all of the calls

It almost feels like it would be nice to have a BaseCharacter function too. Something like
`getCharacter("rahi").forceEquipBy("pc", GlobalRegistry.createItem("ballgag"))`
just a cleaner api. Dunno

PeteTimesSix:
I did notice you're a lot more liberal with introducing new functions than I'd be. Probably I'm just used to languages that have method overloading :) I can rejig it, no problem.

PeteTimesSix:
Something like this, maybe? I've left them in Inventory.gd for now, though, I dunno about having one set of item equipping functions in Inventory and another in BaseCharacter, so if you wanna move it there you'd have a better idea of what you want it to look like.

Alexofp:
Yeah thank you ^^ I will take a better look and merge when not sleepy

Alexofp:
Thankies

PR #36, opened 2023-11-21
rest buff

PeteTimesSix:
And an extremely minor one, a (de)buff towards taking naps.
It might be interesting to also add the multiplier to afterSleeping and adding a debuff to the stocks/bondage gear? I cant imagine napping in stocks is very restful.
That's all Ive got for now :)


Player/Player.gd
```diff
@@ -552,7 +552,8 @@ func afterSleepingInBed():
 func afterRestingInBed(seconds):
 	var _hours = floor(seconds/3600.0)
 	
-	addStamina(_hours * 10)
+	var mult = max(0.0, 1.0 + GM.pc.buffsHolder.getCustom(BuffAttribute.RestEffectiveness))
```
Alexofp, line 555: GM.pc.buffsHolder -> GM.pc.getBuffsHolder()

PeteTimesSix: Fix'd

Alexofp:
Yeah, I think you might as well make it have effect during sleep.
and what I'm thinking.. make it so the sleep restores less pain/stamina depending on how much you already have
What I'm saying is..
Example:
20/100 pain. 50% rest debuff
-sleeping-
10/100 pain
-1 more sleep-
5/100 pain

you get the idea. If you have the rest debuff, you can never fully heal unless you sleep many times in a row x3. Should work like that for the stamina too, you will have to calculate the missing stamina and multiply that

PeteTimesSix:
Like this, maybe? Added it to stocks to try it out, seems to work fine

Alexofp:
I don't think stocks need it to be honest, the stocks scene was balanced without it

PeteTimesSix:
it was actually surprisingly difficult to burn stamina fast enough to test it, so it ends up not having much of an effect - but it seems appropriate. I can remove it if you want?

Alexofp:
Sure, keep it x3

Alexofp:
Thankies

PR #40, opened 2023-12-07
punctuation

Rayforward:
I believe that the ! should not be used in conjunction with the .


Alexofp:
Good catch, my mind just places them automatically I think x3

PR #42, opened 2024-01-11
Persistent items

ranchar1:
This adds a function called `isPersistent` to `itemBase`, which is checked for in `sexEngine` and `inventory` to avoid returning persistent items to the player at the end of sex scenes and to avoid clearing them at the start of sex scenes.

It also adds `onSexStarted` and `onSexEnded` to `itemBase` and calls them at the same time as the similarly named functions for effects and the like in `baseCharacter`.

These changes should make it easier for item logic to be run at the start and end of sex scenes, and for items to be given to NPCs permanently, which is useful for the Portal Expansion Mod that I'm making.

Hopefully, there are no edge cases where this code causes something to break.

### DIFF
--- a/Characters/Character.gd
+++ b/Characters/Character.gd
@@ -223,7 +223,7 @@ func createEquipment():
 	pass
 
 func resetEquipment():
-	inventory.clearEquippedItems()
+	inventory.clearEquippedItemsKeepPersistent()
 	createEquipment()
 	updateNonBattleEffects()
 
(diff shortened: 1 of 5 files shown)


Alexofp:
Heya. Thankies. Little ask:
`clearEquippedItems` should stay as is, clearing all equipped items no matter what. You want to create a separate function and call it `clearEquippedItemsKeepPersistent` and replace the one in the `resetEquipment` with it

Are you sure items need onSexStart/End? Yeah, you said that you need it for the portal panties mod but why exactly. There is no better way?

Alexofp:
Thank you

PR #45, opened 2024-01-21
Added relationship

stb-stb:
- additional stats to hold the relationship with player


Alexofp:
I think that relationship class doesn't mean much without a relationship system x3. I think the new slavery system will fulfill this need for now. Thank you for your efforts though ^^.

PR #46, opened 2024-01-21
Tuned rules for domsex

stb-stb:
- split gagged and oral rules
- you can be more tied to dominate inmates
- for nurses and guards you must also have a stamina
- added requirment to not blindfolded


Alexofp:
I replaced all the requirements with a single ButtonChecks.CanStartSex one, makes it much more clean. And I think it's best to stay consistent about it. So I will close, thank you though ^^

PR #50, opened 2024-01-24
One-handed play support for the cage unlock minigame

PeteTimesSix:
All that work on the mod, and of all the things this one minigame I threw together in an hour continues to be what causes trouble. Here's one last modification to make it not require typing anything and make the hints yet more blatant.

...rats, I was hoping Id manage to sneak this in before the bugfix release.


PeteTimesSix:
Added a few more smol changes.

Alexofp:
Some people complain that they can't figure out how to take off the hypnovisor when they're hypnotized. I understand that it might be the point of it but people still complain x3

Alexofp:
Lemme know if I can merge

Alexofp:
Silence means yes, let's goooo

PeteTimesSix:
Right, sorry. Yeah, hopefully its all good :]

PR #51, opened 2024-01-25
Fixed null pointer

stb-stb:
- caused when non penis person use strap on in scene


Alexofp:
Good catch, thank you ^^. If you know any scenes where that happens, tell me :o

PR #53, opened 2024-01-28
resolved issue with loading game

stb-stb:
- game will not crash while loading save with a scene that no longer exists
- added missing dialog title for confirmation dialog


Alexofp:
Hiding/skipping the fact that the game couldn't find a scene while loading a game sounds very bad, it -will- lead to invalid/unintended game state. Yeah, the current behaviour of just crashing isn't ideal too. The best solution would be showing a message that your save is corrupted and refusing to load that save (returning back to main menu). So, sorry, can't really accept. Thanks for efforts though ^^

PR #55, opened 2024-01-30
Yet more feedback re: hacking minigame incorporated

PeteTimesSix:
Surely. Surely now it's good enough.


PeteTimesSix:
One more thing I thought of. Keeping the SEND and CLEAR buttons around even if there's no current command, so that the other buttons don't jump around when adding parameters.

PeteTimesSix:
Check out Hacknet sometime. Its my favorite take on the idea :]

Alexofp:
Aww, someone beat me to it x3

Alexofp:
oh? I was literally gonna go through the pull requests tommorrow x3

dogboi1120:
i think we need more bodytypes like taur bird etc

PR #56, opened 2024-01-30
Remove duplicate bondage items in fight

stb-stb:
If player has many bondage items of the same type, is listing through many buttons in combat quite annoying, this update offers only one button for each level and information on available quantity.


Alexofp:
Thank you, good change ^^. I will clean it up a bit, I think using a dictionary here would be easier and faster

PR #59, opened 2024-02-05
Default exit in scene

stb-stb:
Add exit scene button if there is no other way to continue


Alexofp:
If that happened, its a sign that something went terribly wrong. So I'd rather for the player to get 'stuck' and report it rather than silently continue but in a broken game state x3

Alexofp:
Yeah, sorry, like I said, better for the player to get stuck rather than getting into an unintended/broken game state

PR #62, opened 2024-03-12
Lock module

stb-stb:
- bondage level can be bigger then 5 (up to player level)
- some kinds of bondage are now lockable
- it is very difficult to remove locked items unless the lock is removed first
- universal key is contraband now and is no more possible to buy it
- new item lock picker can help you to remove the lock
- new item safety cutter can help you to remove some type of bondage stuff
- lockpicker an cutter can be easily destroyed
- removed items can be damaged that reducing their level, especially the cutter usually destroys it
- increased bondage level impact to the difficulty
- rarity modifier make some bondage stuff rare to use by npc
- easier to apply bondage with higher levels in combat
- some bondage perks are more generic and not depend on specific minigame
- even without the proper perk is a chance to get some bondage gear
- instant escape is more easier escape, instant is only simple stuff
- unified minigames flow
- new match dice minigame 
- new minigame for development (if you are in dev mode you get it)
- removed info about game effectiveness (total progress is still here)
- new text processing for struggle allows better text compose


Alexofp:
This is too much, I'm sorry, you -have- to split it. One feature at a time pls :o

Alexofp:
Thank you for your efforts. I'm sorry if I seemed mean x3

PR #63, opened 2024-03-12
Restraint over level 5

stb-stb:
- generating restraints level in accordance with the character's level
- restrain level can be over 5

### DIFF
--- a/Inventory/ItemBase.gd
+++ b/Inventory/ItemBase.gd
@@ -384,11 +384,12 @@ func setRestraintLevel(newlevel):
 	if(restraintData != null):
 		restraintData.setLevel(newlevel)
 
-func calculateBestRestraintLevel():
-	#if(GM.pc != null):
-	#	return GM.pc.calculateBestRestraintLevel()
-	#else:
-	return RNG.randi_range(1, 5)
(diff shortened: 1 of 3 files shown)


Fox2Code:
I sure would like higher level on restraints, but I don't think scaling it with level is a good idea.

I would rather it being tied to other elements, like BDSM skill level instead?

(Personally I would also want to be able to upgrade restrains, but this is out of scope of this PR)

Fox2Code:
Note: I'm aware this can always be changed again in the future.

Fox2Code:
A restrain losing level is a bit problematic, maybe a new "condition" property would be better, a "condition" is how well an item is maintained, and would make it easier to escape if it's in bad condition?

If it's 0% it get loose (auto unequipped) and cannot be put back on?

Alexofp:
The game is balanced around restraints with levels 1-5, I don't see the reason behind adding more :o. I would honestly keep stuff as is, some scenes already use high-level restraints. I don't mind high-level restraints but only for special scenes, you know? Random stuff should be fairly easy to struggle out of, that's my reasoning at least

Alexofp:
I think the smartlocks helps

PR #64, opened 2024-03-12
Lm rarity

stb-stb:
- items used by npc's may contain a rarity value that determines how rare their use is
- hypnovisor is now less common than common blindfold


Alexofp:
There is already a function that makes certain restraints less used by NPCs. `getAIForceItemWeight()`

I updated it so the hypnovisor is now more likely to be used by NPCs who have the hypnosis fetish, should make more sense and lower their amount hopefully x3. Closing as its not needed. Thank you for bringing it up though

PR #66, opened 2024-03-12
Lm refactor sq

stb-stb:
- bondage gear refactoring
- on all things that inherit from basic bondage can be use struggle, lockpick and cut action.
- flow for process and calculations is unified so that it is possible to replace only relevant small parts
- by default all calculations are based on level and perks
- advanced text composition allows write reusable code
- every acrion support positive and negative numbers from the minigame (from great success to fatal fail)
- unified interface to minigames - compatibility with clicking minigame preserved

Warning this is compatibility breaking change! Struggling has different return type some mods won't work without update
This version needs test minigame already merged

### DIFF
--- a/Inventory/Items/BDSM/HypnovisorMk1.gd
+++ b/Inventory/Items/BDSM/HypnovisorMk1.gd
@@ -51,7 +51,7 @@ func isRestraint():
 
 func generateRestraintData():
 	restraintData = RestraintHypnovisor.new()
-	restraintData.setLevel(5)
+	restraintData.setLevel(calculateBestRestraintLevel())
 	
 func getForcedOnMessage(isPlayer = true):
 	if(isPlayer):
--- a/Game/SexEngine/SexActivity/DomBondage.gd
+++ b/Game/SexEngine/SexActivity/DomBondage.gd
@@ -210,7 +210,7 @@ func processTurn():
 		
 		if(dom.isPlayer()):
 			dom.getInventory().removeItem(item)
-		if item.isRestraint() && item.getRestraintData().canBeLocked() && RNG.chance(66):
+		if item.isRestraint() && item.getRestraintData().canBeLocked() && RNG.chance(33):
 			item.getRestraintData().lockIt()
 		sub.getInventory().forceEquipByStoreOtherUnlessRestraint(item, dom)
 		#sub.getBuffsHolder().calculateBuffs()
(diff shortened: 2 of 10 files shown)


Inventory/Items/BDSM/HypnovisorMk1.gd
```diff
@@ -51,7 +51,7 @@ func isRestraint():
 
 func generateRestraintData():
 	restraintData = RestraintHypnovisor.new()
-	restraintData.setLevel(5)
+	restraintData.setLevel(calculateBestRestraintLevel())
```
Fox2Code, line 54: This item is sold by an NPC in the game, maybe the restrain level should stay a constant?

Fox2Code: Hypnovisor = dropped item by common NPC's
Hypnovisor**Mk1** = a special item only unlocked and sold after a specific quest.
The **Mk1** variant is **never** used by common NPC's. (Until it's changed at least)

But for selling or specific quest events, I agree a way to force specific level loot could be nice.

Alexofp:
I looked through it. Sadly I don't think I can accept it, I'm sorry ;-; It seems like it took effort but it doesn't fit my vision I guess
I don't like restraint keys being replaced with tools that require minigames. If you're using some resource, that should be enough by itself..
Struggling code is already a mess but I think the refactor only makes it worse x3. Maybe one day I will rewrite and simplify everything..
And it's still too much code I think.. Please focus on something simple and easy rather than trying to reimplement a core mechanic of the game.. Sorry if I sound mean, I don't want to seem mean, I'm just very tired x3

Alexofp:
You were certainly on the right path. I do have similliar ideas, it's just that they're different enough from your pull request that it would be easier for me to implement them with my current codebase..
The biggest issue that everyone seems to have is that the struggling becomes too easy as you level up.

My ideas to fight that (don't read if you don't want spoilers x3):
1. Some restraints that can only be unlocked ONLY through a use of a very limited resource. Example: Extra tight blindfold that -needs- to be cut through with a shiv or something sharp (breaking the shiv). No shiv = can't remove it. Just a single example
2. Restraints with smart locks. Again, can't struggle out of them. The smart lock requires you to do certain things to be unlocked (like it can track how many times you were fucked and unlock after a certain amount, as an example). These ones should also be unlockable with a restraint key probably in case you really don't want to do what the restraint asks you to do
3. Probably the hardest one to implement. A npc locks a restraint onto you and keeps the key. Can't struggle out of these, it has to be unlocked with that key. To get the key you gotta interact with the npc (would require a lot of writing)

That's how I would battle the problems of the restraints getting too easy to escape from. I might start implementing my ideas at some point, when I have mood and time x3

PR #68, opened 2024-03-12
new dice minigame

stb-stb:
New minigame based on catching dice with same number, more numbers and bigger value means bigger score


Alexofp:
It's a cute minigame, I think it would fit somewhere else though. Like, inmates would totally play some dice game against each other
But dice and bdsm restraints don't really fit, I don't think.. Hope that makes sense?

PR #75, opened 2024-05-01
get a roomName

Rayforward:
No roomNme so in game it shows "MainHallRoom48".I wonder if you would like use another roomName……


Alexofp:
Good catch thank you ^^

PR #76, opened 2024-05-03
Use roomName = "Elevator"

Rayforward:
Since elevators on other floors are ##  roomName = "Elevator"    ##, I thought there should be a similar name here. Or you can change the other three places to ##   roomName = "Main elevator"  ##.XD


Alexofp:
Yeah having the same name everywhere makes sense, thankies

PR #79, opened 2024-05-04
fixes crashing when argument size is not equal to what command expects

### LINKED ISSUE #78: Console command, when missing arguments command_entry's paramCount index does not exist
When using a console command that requires argument, it crashes due to trying to call non existing index of `"paramCount"`
https://github.com/Alexofp/BDCC/blob/main/UI/DebugUI/ConsoleSingleton.gd#L77

CanInBad:
Fixes #78


Alexofp:
Heh, thank you

PR #80, opened 2024-05-06
Add spinbox for stat points

PosionFox:
this pull request adds a spinbox that lets you set the amount of stat points to a given primary stat


Alexofp:
I understand that being able to add more than one point at a time would be a handy feature but I think that adding a spinbox near each stat makes it very confusing for new players. They would set the spinbox to the desired value.. and wonder why the apply button is still gray. I'd maybe just add a +5, +10 buttons and rename the current + into +1

Alexofp:
Thank you for addressing this though, hope you will understand what I mean x3

PosionFox:
okay i understand

PR #83, opened 2024-05-19
Adding xtoys.app webhook functionality for interactions with toys.

dereve-prog:
The webhook is generic enough to use with any webserver accepting JSON, but was created for xtoys.app.

How to test:
- Follow Misc/Webhook.md instructions, should work even without a bluetooth connection to the toys.

### DIFF
--- a/Game/MainScene.gd
+++ b/Game/MainScene.gd
@@ -1033,6 +1033,13 @@ func _on_Player_holePinafullyStretched(bodypart, _who):
 		addMessage("OW! Your pussy [b]hurts[/b]..")
 	if(bodypart == BodypartSlot.Anus):
 		addMessage("OW! Your anus [b]hurts[/b]..")
+	callPainWebhook(bodypart)
+
+func callPainWebhook(bodypart):
+	var jsonStr = ""
+	if(OPTIONS.getWebhookPainEnabled()):
+		jsonStr= '{"action": "painstretch","bodypart": "' + str(bodypart) + '"}'
+		$HTTPRequest.request(OPTIONS.getWebhookURL(), ["Content-Type: application/json"], true, HTTPClient.METHOD_POST , jsonStr)
 
 func _on_Player_gotWoundedBy(_who):
 	addMessage("OW! That [b]really[/b] hurt..")
--- a/Game/MainScene.tscn
+++ b/Game/MainScene.tscn
@@ -28,6 +28,8 @@ script = ExtResource( 2 )
 
 [node name="DynamicCharacters" type="Node" parent="."]
 
+[node name="HTTPRequest" type="HTTPRequest" parent="."]
+
 [connection signal="onDevComButton" from="GameUI" to="." method="_on_GameUI_onDevComButton"]
 [connection signal="on_option_button" from="GameUI" to="." method="_on_GameUI_on_option_button"]
 [connection signal="on_rollback_button" from="GameUI" to="." method="_on_GameUI_on_rollback_button"]
--- a/Game/Options/GlobalOptions.gd
+++ b/Game/Options/GlobalOptions.gd
@@ -76,6 +76,9 @@ var visibleWritings:bool = true
 var autosaveEnabled = true
 
 var genderNamesOverrides = {}
+var webhookURL = "https://webhook.xtoys.app/abcrandomidentifier"
+var webhookStatsEnabled = false
+var webhookPainEnabled = false
 
 func resetToDefaults():
 	fetchNewRelease = true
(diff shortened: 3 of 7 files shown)


Alexofp:
Pretty cool but I can't really maintain it since I don't use that website and the setup seems to be a bit tricky x3. Ideally you'd somehow turn it into a mod, but that's probably gonna be tricky too..

dereve-prog:
Is there a way to expose stats updates to mods? Some kind of an external UI mod? I've seen mostly additional .gd files or .tscn files in mods, I'm not sure how extendable hooking would work in Godot.

Alexofp:
Its possible to add a few game-related callbacks to the Module class maybe. And mods can add their own modules easily, they are objects that always exist loaded in memory

dogboi1120:
i had an idea: species: kangaroo

Alexofp:
I'm working on sex toy integration in the dev branch. It's still work-in-progress but most of it is done. It supports 2 backends: buttplug.io and xtoys.app

I wonder if the xtoys backend is intuitive-enough to use

Zsar:
A comment on that, if I may, because this is an issue I have seen in a number of other Buttplug.io integrations:

(Obviously) most physical devices will not support all the features that Buttplug.io can route. But that means that the game must have some logic to replace/reroute events for unsupported features with/to supported features, lest "holes" without / with less-than-intended activity open up.

At least as far as I can see, there is no such feature native to the Intiface.

Alexofp:
Problem is that I don't have toys with unique features and there doesn't seem to be a way to simulate them x3
If you have strokers or fuck machines, your best bet would be connecting them using xtoys app, that app seems to support every toy on the planet x3

Zsar:
Ah, sorry, I think you understood the opposite of what I meant. I'll try again:

Let's say you add toy support to the egg plant scene. It would penetrate and inflate the character, so you query `device.hasOutput("Oscillate")` and `device.hasOutput("Constrict")`.

Let's say the only device connected is a vibrator, so both calls return `false`.

If you just send the command anyway, (as far as I know), devices will simply silently ignore the unsupported part and only do whatever part of the command they _can_ do.

So, in this particular case nothing would happen at all.

_Unless_ you query the device's _other_ capabilities and select ones it actually has as fallback.

What this means in practice is creating a directed graph of fallbacks over the nodes of [this list](https://buttplug.io/docs/dev-guide/architecture/client-device-in-depth#output-types).

The messy part is that some things just cannot be sensibly substituted. E.g.
- If a device does not support `Temperature` then it might be best to indeed drop it.
(Though maybe it could indeed be substituted with `Constrict` or with `Vibrate`, if these were not already in use simultaneously...)
- If a device cannot `Constrict` then `Position` or `HwPositionWithDuration` _may be_ adequate substitutions, but almost certainly not with the same values.
(Though it may perhaps be best to not touch these two at all... - I cannot even imagine how one would arrive at sensible values for them, if not by testing _exactly the device at hand_.)

I do not think it is necessary to complete this feature at the first release, but it would be sensible to put the necessary code in place to expand upon it later on. (Say, via Merge Request from people with the relevant hardware.)
E.g. some easy substitutions ought to be

- `Rotate` / `RotationWithDirection` => `Vibrate`
- `Oscillate` => `Vibrate`
- `Vibrate` => `Oscillate`

and one particular issue to be considered right away might be "What if I already have vibration but I also need to substitute Oscillation somehow?" - it would be necessary to scale down the intended values as to make room for adding the intended vibration value and the substituted oscillation value together without going out of bounds.

... I guess it might be sensible to, instead of figuring out sensible transformations in code, support a configuration file, where users can define their own mappings. But that would severely limit this feature's accessibility, perhaps even more than simply requiring toys with specific sets of features and not mapping anything.

(There are not really "unique features" for toys as a general category, are there?

- If you have a vibrator, of course it will be able to vibrate.
- If you have a stroker, of course it will be able to oscillate.

But the other way around, the capability would be unusual.

... Well, currently I guess `Temperature` is an unusual capability, no matter the kind of toy - I have certainly been searching for years to find some and whenever I find one that _alledgedly_ heats up, there are always plenty of reviews that state it actually does not.)

Zsar:
Err... you do realise that [Buttplug.io is already compatible with most Lovense products](https://iostindex.com/?filter0ButtplugSupport=4&filter1Brand=Lovense) and will most likely become more compatible with the missing ones on it's own?

Integrating a Lovense backend in addition to a Buttplug.io backend does seem rather wasteful.

(FWIW: I only have one Lovense product lying around myself, the Max 2, and that for that one Buttplug.io support is _straight up better_ than Lovense's own app. Maybe there are some very popular ones that are currently unsupported? - But even then, the more popular they are, the more likely Buttplug.io will support them in the future.)

Alexofp:
yeah I know. I just did it because I felt like

Alexofp:
This PR got superseded by the new sex toy integration. I did look at this PR while coding it so big thank you <3

PR #89, opened 2024-06-26
Spelling, grammar and punctuation

Disillusioned-Gazer:
I'm a programmer, but for now all I can contribute to this project are nitpicks like these that I noticed while playing.


Alexofp:
Thank you very much ^^

PR #106, opened 2024-08-29
Fixes console/cout spam when deciding max value in getRoughLitterEstimateString

CanInBad:
This PR fixes console or cout print spam from RNG randi_range() that's bugging me on my playthrough.

Here is my issue, in my playthrough I noticed this strange spamming inside the console every time I walk around:
`randi_range() from is higher than to. From = 36 To = 34`
(36 is from calculation of then line 310 and 34 being trueValue)

I have tracked down the problem for me, I haven't test on other saves but it should work fine.

Here's the [save](https://files.catbox.moe/b3zd5w.save)
and you at least require [Hypertus](https://github.com/CanInBad/Hypertus) to load into the save then try walking around


Alexofp:
Yeah, seems good, thank you ^^

PR #107, opened 2024-09-05
Fixed crashing due to null pointer in getting species name

CanInBad:
This is an issue because if the species does not exist anymore (eg. removing species mod and looking at nursey database)


Alexofp:
Looks good, thankies ^^

PR #111, opened 2024-09-20
part orientation

CanInBad:
I'm pretty sure I got all the edge cases and I sure hope there is no weird doll scale in the future.

if you have anything you want to changed you can edit them or discuss it with me

### DIFF
--- a/Player/Player3D/Parts/PartOrient.gd
+++ b/Player/Player3D/Parts/PartOrient.gd
@@ -1,0 +1,9 @@
+extends Spatial
+class_name PartOrient
+
+export(bool) var faceRight = false 
+# if false, will only be visible when facing left (eg default player pose)
+# if true, will only be visible when facing right
+
+func getOrient():
+	return faceRight
--- a/Player/Player3D/Parts/PartOrient.tscn
+++ b/Player/Player3D/Parts/PartOrient.tscn
@@ -1,0 +1,6 @@
+[gd_scene load_steps=2 format=2]
+
+[ext_resource path="res://Player/Player3D/Parts/PartOrient.gd" type="Script" id=1]
+
+[node name="PartOrient" type="Spatial"]
+script = ExtResource( 1 )
--- a/Player/Player3D/Parts/PartOrientPicker.gd
+++ b/Player/Player3D/Parts/PartOrientPicker.gd
@@ -3,8 +3,6 @@ class_name PartOrientPicker
 
 export(bool) var invert = false
 
-var _callcount:int = 0
-
 func _ready():
 	var theNode = get_node("..")
 	var theDoll:Spatial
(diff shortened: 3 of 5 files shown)


Alexofp:
Like I said before, just checking it once in ready isn't enough

Sometimes (very rarely), the game can do this.

`playAnimation(StageScene.Duo, "stand", {npc="nova"})`

All is good so far. But then it can do this when the player picks something

`playAnimation(StageScene.Duo, "stand", {npc="nova", flipNPC=true})`

This flips the npc without recreating the scene.. so the ready function won't be called and stuff.
How to fix it? I dunno x3. Apart from adding a `_proccess` function and checking if our global scale has changed each frame..

CanInBad:
>All is good so far. But then it can do this when the player picks something  
>`playAnimation(StageScene.Duo, "stand", {npc="nova", flipNPC=true})`

I fixed that by looping enough until it reaches `BaseStageScene3D` type node

https://github.com/Alexofp/BDCC/blob/90cb98f4a6cbff64e415049f01c75fb33deb0086/Player/Player3D/Parts/PartOrientPicker.gd#L10-L17

and checking if the node is actually `BaseStageScene3D`, then getting `getVarOptions()` to check if it has `flipNPC` option or not
if yes it will wait until there is idle frame, then recheck if it has flipped again

https://github.com/Alexofp/BDCC/blob/90cb98f4a6cbff64e415049f01c75fb33deb0086/Player/Player3D/Parts/PartOrientPicker.gd#L28-L31

unless you're planning to make the doll flip in real time then this will do for now

Alexofp:
Sometimes the doll is flipped long after it was created, that's what I'm trying to say x3

SCENE STARTED
PLAYING ANIMATION WITH NOVA
PLAYER PICKED AN OPTION
PLAYING THE SAME ANIMATION WITH NOVA (FLIPPED)

the last line won't cause the update.. because the animation is the same.. so nothing is re-created. I hope that makes sense x3

CanInBad:
ill return with another implementation

Alexofp:
aw, I'm sorry, thank youuuuu

CanInBad:
I didn't meant to close the pull request, i was eepy

CanInBad:
I sure hope I got the cases you mentioned, I have made a scene to test the feature out here, just extract it in root, equip the test clothing and run the test scene id `partOrientTester`  
[PartOrientTester.zip](https://github.com/user-attachments/files/17083339/PartOrientTester.zip)

Alexofp:
It seems very hacky even now but its probably fine for now. I dunno if people even need it that much but sure x3. Thank you

Alexofp:
https://github.com/Alexofp/BDCC/commit/69d21d98e8dd9729c3bb6d9eba8315cf7b1c91e5

I coded an easier solution. You don't need any custom part pickers. You just use a normal one with a 'facing' state. And possible states are 'left' and 'right'. Doesn't check it every frame too so it should work faster

PR #120, opened 2024-10-26
Fix grammar in Tavi module (borned -> born)

Friskygote:
I dislike making such small pull requests but it is what it is


Alexofp:
Thank youuuu, borned sounds so silly x3

PR #124, opened 2024-11-10
Input icons and better controller support

halotroop2288:
- The in-game option buttons have had their hotkeys changed to Input Map actions to support multiple bindings
- Text labels for inputs have been replaced with icons
- Controller bindings now exist for more actions to make navigation easier
- Previous Page and Next Page buttons now show their binding

<details><summary>Screenshots (<b>Warning: Adult content</b>)</summary>
<p>

image
image

</p>
</details>


Alexofp:
Is there a big demand for controller support?.. It's a lot of images, just afraid that it will bloat the game size for people who will never use controllers..

export_presets.cfg
halotroop2288, line 1: Note to self: This was not meant to be part of this PR.

halotroop2288:
> Is there a big demand for controller support?.. It's a lot of images, just afraid that it will bloat the game size for people who will never use controllers..

It's only like 10MB. Actual content will take up way more than that over time, considering how many sprites are in the game. This helps improve the UX for even keyboard users, it's well worth it.

pl23:
I'm only saying this because an Xbox controller layout is more commonly used why not add x input icons? You could possibly have it built on a toggle based on which controller is detected or just have it be a setting option that you can swap between the two

halotroop2288:
That's what it already does.
It detects your controller hardware based on Godot's pad metadata.

PR #128, opened 2024-12-20
Add positional argument to ParadedOnALeashScene to specify player's pose

Friskygote:
Hello there!

I've been making a mod for a month or so, and decided to submit this little PR for adding optional pose for player in ParadedOnALeashScene. I think game/mod devs would benefit from additional argument there as it allows for a bit more petplay options of player walking around on their knees instead (personally using "crawl" action in few places for this scene).


Alexofp:
Looks good, thank you ^^

PR #129, opened 2024-12-20
Fixing targeted object in doSpitCumIntoHoleDom function

Friskygote:
Seems like simple copy-paste error. Text refers to wrong person in the dialogue resulting in nasty ERROR:NO_VAGINA error on sex.


Alexofp:
Good catch, thank youuu

PR #135, opened 2025-01-04
Adds a new option for changing how overstimulation scales

PosionFox:
I would still like some guidance as i am not sure if the way the scaling is being applied the correct way


Alexofp:
I think I'd rather balance it out myself until everyone is happy-ish rather than letting people trial-and-error a good value for themselves..

PosionFox:
that's fair

PR #137, opened 2025-01-04
Added new option for adjusting block catcher panel size

PosionFox:
I just recently discovered this thing existed and had no idea you could drag code blocks above others, it wasn't intuitive at all and mobile users had it rough, so i changed the default size of the node and made the size a setting, also added a very thin outline to highlight it

### DIFF
--- a/Game/Datapacks/UI/CrotchCode/VisualSlots/BlockCatcherPanelSmall.gd
+++ b/Game/Datapacks/UI/CrotchCode/VisualSlots/BlockCatcherPanelSmall.gd
@@ -12,6 +12,6 @@ func drop_data(_position, _data):
 	emit_signal("onBlockDraggedOnto", _data, dropIndex)
 
 func _ready():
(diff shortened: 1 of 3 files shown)


PosionFox:
imagen_2025-01-04_015050354

it looks like this with 16 pixels of height

Alexofp:
I'm not sure I like how it looks.. the bright color is too distracting. With many blocks it would look even worse probably.. Not sure, maybe keep the colors subtle is the way.. but I think the old solution already had dim colors x3

PosionFox:
yeah looks good without bordering
Captura de pantalla 2025-01-04 023020
here's how it looks like with 8 pixels, might be the sweet spot for most but i prefer 16
Captura de pantalla 2025-01-04 023049
4 pixels is still a thing in the settings btw

Alexofp:
yeah that looks better

Alexofp:
Thank youuuu

PR #139, opened 2025-01-14
Fixed implementation of getSharedChildrenAmountFatherMother method

Friskygote:
This PR aims to fix two bugs and add a few more type hints and explanations inside ChildSystem.gd

## Problem
getSharedChildrenAmountFatherMother function used for datapacks and nowhere else in the project claims to return "Child amount" with specified Father and a Mother. 
However, with current implementation it returns children with given father OR with given mother and it improperly asks for archived children.

I don't believe that was the intention of that function, considering the fact that the part adding amount of archived children would work differently.

## Solution
Instead of using logical or  (||) use logical and (&&) when checking whether a child has given father and mother. 
Also switch the order of arguments to archive function to match the function's expectations (first mother, then father, inverted in how it's in the datapack call/getSharedChildrenAmountFatherMother call).

Also added more type hints as well as some short descriptions of function functionalities.

### DIFF
--- a/Game/Pregnancy/ChildSystem.gd
+++ b/Game/Pregnancy/ChildSystem.gd
@@ -72,7 +72,7 @@ func getSharedChildrenAmountFatherMother(firstCharID, secondCharID) -> int:
 	for child in children:
 		if((child.getFatherID() == firstCharID) && (child.getMotherID() == secondCharID)):
 			result += 1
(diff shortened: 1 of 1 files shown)


Alexofp:
Yeah, I think you did it correctly x3. I dunno, it was very exhausting making all those codeblocks, probably more fuckups somewhere else x3.

Thank you!

PR #145, opened 2025-02-26
Fix unlikely actions within interactions not being filtered out

keerifox:
Sometimes pawns will perform a "justleave" action with 1% probability, even if the threshold/cutout was intended to be set to 10%.


Alexofp:
Good catch, thankies

PR #148, opened 2025-03-05
Add ability to refuse help with restraints, stocks and the slutwall

keerifox:
Pawns always assume their help with restraints is wanted. With the changes in this PR, offering to help someone with restraints prompts them on whether they wish to be helped, with the name of the restraint being displayed in the event log.

Stocks/Slutwall already has an ability to prevent help, but not before the initial attempt was already made. It's now possible to stop them in anticipation, featuring some dialogue lines added for extra flavor.

NPC's may refuse being helped if they are into bondage and the helping character doesn't have dommyness advantage over them. They will not refuse help if they shouted for help at least once.

\-

Fixes other restraints being removed when someone is being helped to struggle out of stocks or the slutwall, by giving priority to the latter.

### DIFF
--- a/Game/BaseCharacter.gd
+++ b/Game/BaseCharacter.gd
@@ -3018,39 +3018,83 @@ func hasSomethingToStruggleOutOf() -> bool:
 		return true
 	return false
 
-func doStruggleOutOfRestraints(isScared:bool = false, addStats:bool = true, customActor=null, damageMult:float = 1.0) -> Dictionary:
+func getRestraintsToStruggleOutOf() -> Dictionary:
 	var possible = []
 	var trivial = []
-	
-	var whoStruggles = self
-	if(customActor != null):
-		whoStruggles = customActor
-	
+
 	for item in getInventory().getEquppedRestraints():
-		var restraintData: RestraintData = item.getRestraintData()
-		
+		var restraintData:RestraintData = item.getRestraintData()
+
 		if(restraintData == null || !restraintData.canStruggleFinal() || !restraintData.shouldStruggle()):
 			continue
-		
+
 		if(!restraintData.shouldDoStruggleMinigame(self)):
 			trivial.append(item)
 		else:
 			possible.append(item)
-	
-	var pickedItem
-	var minigameResult:MinigameResult
+
+	var restraintsToStruggleOutOf = {
+		possible = possible,
+		trivial = trivial,
+	}
+
+	return restraintsToStruggleOutOf
+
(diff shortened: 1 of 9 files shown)


keerifox:
Absolutely no worries about taking the time (the daily progress on `dev` branch is just immense), or the difficulty understanding the code ^^ Most of it is not so much new as it is moved around.. I can try to summarize the core points:

## BaseCharacter.gd

`doStruggleOutOfRestraints` would struggle out of a random restraint, and we needed to know the exact restraint that's about to be struggled off next (so we know what we're agreeing to..) so i split its logic into:
1) `getRestraintsToStruggleOutOf` which returns all of them
2) `getNextRestraintToStruggleOutOf` which will no matter how many times you call it (with the same hash), should always return the same restraint, until it's no longer worn<sup>*</sup>

<sup>* – or the list of restraints changes in any way. but in the context we're using this, it should only happen when the restraint is successfully removed</sup>

\-

## InStocks.gd / InSlutwall.gd

The changes in `InStocks.gd` and `InSlutwall.gd` are identical – instead of immediately being helped with restraints, the inmate can accept or refuse help in `refusable_save_using_<stamina/key>`.

Sorry for the scary diff there, I've moved already existing code into `saveUsing<Stamina/Key>` functions since there are now 2 states that reuse the code. After you agreed to be helped, v0.1.7fix1 already has an option "Stop them", so the new behavior only applies to 1st attempt to help someone, 2nd+ it works as it used to.

I've also changed the wording of "leaves, no longer trying to save you" when someone has.. like.. approached you, didn't even try anything, just left.

\-

## HelpingWithRestraints.gd

This interaction can be started when asking someone to help take your restraints off, but it can also be `"reacterStarted"`, that is.. someone walks up to you and they're like "hey i'm helping you take this off".

The changes in this are more or less straightfoward, the game now lists the name of the restraint that will be removed, and pc/pawn can accept or refuse help with it.

Once the restraint is taken off, if there are other restraints and have stamina to help more, the interaction will go back to `init_eager_to_help` state and ask if pc/pawn wants help with that other restraint.

<br>

\\ \\

<br>

I've tested it on v0.1.7fix1 and it seemed fairly stable (no relevant code has changed since, I believe).

I think it would improve on the variety a bit, having a possibility that the pawn will not want your help, but it would also be nice to have the ability to refuse the help yourself if you want to keep a certain restraint.

Fox2Code:
I checked the code of this PR, it's seems fine to me, but I could have missed a small detail.

Alexofp:
Yolo x3 Thank you very much

PR #156, opened 2025-03-25
Allow quicker switching between skin patterns, indicate the selected option

keerifox:
when I used the character creator for the first time, it took quite a bit of time to try out every possible pattern, and there was no indication of which one was selected, so I propose a smol quality-of-life tweak that doesn't change a lot

\-

without PR:

https://github.com/user-attachments/assets/37152fe8-aab0-4d4b-9975-a4c258cf9422

with PR:

https://github.com/user-attachments/assets/5ad5af3c-22ad-48ec-b6f3-f7cc897a7bd9

\-

I couldn't find a sensible way to prevent skips to page 1, that can be done outside of the PR if desired \~w\~

### DIFF
--- a/Scenes/ChangeSkinScene.gd
+++ b/Scenes/ChangeSkinScene.gd
@@ -88,7 +88,10 @@ func _run():
 		addButton("Back", "Go back", "basemenu")
 		for skinID in GlobalRegistry.getSkinsAllKeys():
 			var theSkin = GlobalRegistry.getSkin(skinID)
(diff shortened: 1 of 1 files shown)


Alexofp:
Thank you, I will try to add a thingie to save the current page

Alexofp:
Did it in https://github.com/Alexofp/BDCC/commit/fb8fa2d0d47c4f82c7f0f81a7c489b310abad0e5 works really well. Might be worth it to do this in other places x3

PR #157, opened 2025-03-27
[Feat] add a new bing translator and a setting for translator order

CKRainbow:
Add a new bing translator and a setting for translator order. Mainly because it is hard to access google translation for some region.
image

The bing translator works well, to chinese.
image

And maybe you want to change few things of UI and logic to make it consistent with the original code and feel free to.


Alexofp:
Looks good at a first glance. Thank you <3. Sorry, it's taking me a while to go through all the tasks x3

CKRainbow:
thx a lot! and I will figure out how to fix the deepL if i have time.

PR #160, opened 2025-04-25
Fix typo in sleeping scene

Friskygote:
Honestly, can't believe this typo survived for so long in such often reoccurring scene lol


Alexofp:
Yeah x3. Guess there is no reason to read into it too much

Thank you

PR #161, opened 2025-04-27
fix: pose descriptor mixup and press/presses

keerifox:
been there done that \~w\~


Alexofp:
Good catches, thank youuuu

PR #164, opened 2025-05-02
Skip dialogue when applying lube (or a modded drug/lube)

keerifox:
when applying lube, dom would say one of those things

https://github.com/Alexofp/BDCC/blob/f1c6a2bb8fabf22601a467249a74bca977022b7a/Game/SexEngine/SexVoice.gd#L723-L730

could write some generic phrases like "Let's get you ready.." that would still make sense with potential modded items but i'm not creative enough right now x3

so this change just omits the dialogue for now


Alexofp:
Yeah, makes sense, thank youuuu

PR #165, opened 2025-05-02
Prevent most PunishInteraction actions with restraints on arms or paws

keerifox:
another 0.1.7 thing that i haven't paid close attention to, but you can clip a leash to someone's collar, restrain them, choke them out etc. all with arms tied and/or paws blocked, which is veeery flexible but seems unintentional \~w\~


Alexofp:
There are a looooot of places where restraints aren't checked x3. I gave up even trying to adjust scenes, the amount of permutations of restraints is just too high x3. But sure, thank youuu

keerifox:
yeah that's trueee, at least this interaction is pretty common

PR #166, opened 2025-05-02
"temporary takes out a thing out of" -> "temporarily retrieves a thing out of"

keerifox:
i'm not 100% sure on the wording here, but ideally it should probably be "temporarily retrieves the thing out" / "temporarily removes the thing from", just to avoid saying "out" twice


Alexofp:
english hard thank

PR #167, opened 2025-05-03
hypnokink: First action always obeys if pc ReallyLikes/Loves being a subject

keerifox:
i like the idea of needing to look for the right option when trying to resist, but in my opinion obeying shouldn't require the effort to make sure you don't resist by accident \~w\~

hard to explain why i tied it to fetish interest, feels more personalized for those in a specific mindset

changelog:
added "head empty" gameplay


Alexofp:
Sure. Thank you

PR #168, opened 2025-05-03
Fix "namehere is in a AloneInteraction interaction"

keerifox:
i'll miss walking up to someone just to learn they're "in a AloneInteraction interaction" but it's probably worth a fix x3


Alexofp:
If you see that, it means there is a bug probably x3. But sure, thank you

PR #173, opened 2025-05-03
Fix not being able to milk self if the pump is already worn

keerifox:
pump not pumping part 2


Alexofp:
Makes sense, thank you

PR #174, opened 2025-05-04
hypnokink: Fix selecting "OBEY" often resulting in a resisting action

keerifox:
selecting the "OBEY" action would still keep pc's resistance at whatever % it was, this patch sets it to 0% for the duration of OBEY, then restores it to the previous value (to keep it acting as intermittent loss of control that it seems to have been intended as)

i initially started with the blocklist of actions and then realized that overriding pc's resistance solves ~80% of actions. though it's still preferred to have both, just in case (and for the remaining 20% actions)

\-

there's also a bit of code that makes "OBEY" force pc to accept/swallow drugs (over any other action), even if the score for accept/swallow action is <= 0. they however won't do that if need to *finish* during that turn.

### DIFF
--- a/Game/SexEngine/SexEngine.gd
+++ b/Game/SexEngine/SexEngine.gd
@@ -745,6 +745,35 @@ func processAIActions(isDom = true, playerIsHypnotized = false):
 			var importantScores = []
 			#var importantScore = 0.0
 			
+			if(playerIsHypnotized && (personID == "pc")):
+				# it's very likely that the scores for most of these are already at <= 0 as a result
+				# of setting subInfo.resistance to 0.0 for the duration of OBEY turn, but we can
+				# exclude them here too, just to be safe & to allow adding exceptions if needed
+				var resistingActivityIDs = ["SubBegToPutOnACondom"]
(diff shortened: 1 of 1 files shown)


Alexofp:
Feels very hacky, especially checking a hardcoded list of action ids. I will try to do it differently, sorry x3. But thank you, it's a good idea

keerifox:
@Alexofp felt a little bloaty to add an extra field to every action to mark it as impossible to auto-select when choosing OBEY, since it's only used in this one part of codebase, but you or me could do that instead ^^

Alexofp:
That's how I did it
https://github.com/Alexofp/BDCC/commit/eda066012957d2a4742629bf50ec06e32b7e0bb2
I didn't have to touch the resist actions because they all depend on the getResistScore() and that one now always returns 0 if you're forced to obey

keerifox:
@Alexofp yup that should also be good enough, I'll try to cover some cases it might miss

Alexofp:
Not today, sorry x3. I'm sure it will be fine x3

keerifox:
@Alexofp hehee it's insignificant stuff so even if takes months that's okay

PR #176, opened 2025-05-04
Reversible and slightly more precise zoom

keerifox:
very smol annoyance but i didn't know pressing the middle mouse button resets the zoom so whenever i tried to restore previous zoom it would be messed up

without change it goes like: 1.00 -> 0.90 -> 0.99 -> 0.89 -> 0.98

with change it goes: 1.00 -> 0.90 -> 1.00

\-

also compared to the minimap, 0.90 for the character model view seems too sharp of a change, so i suggest 0.97 (which is still pretty strong)

### DIFF
--- a/Game/Datapacks/UI/CrotchCode/UI/AnimPickerWidget.gd
+++ b/Game/Datapacks/UI/CrotchCode/UI/AnimPickerWidget.gd
@@ -294,9 +294,9 @@ func a_gui_input(event: InputEvent):
 	
 	if event is InputEventMouseButton:
 		if(event.button_index == BUTTON_WHEEL_UP):
-			camera3d.size *= 0.9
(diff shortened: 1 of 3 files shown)


Alexofp:
I kinda liked that it wasn't precise x3. That way you could dial it up to a perfect value by scrolling up+down

keerifox:
ohh interesting, i think it's better to reduce the increments instead, i'd assume many players would expect back-n-forth zoom to restore to default value, but that's something i didn't consider

maybe Alt or some other key could make it a very slight change

Alexofp:
Yeah, sorry, I kinda prefer the current behaviour. If they want the default value, the middle click already resets the zoom. Sorry x3

keerifox:
no worrs, up to you

PR #177, opened 2025-05-04
"OBEY" accepts/swallows pill

keerifox:
this was part of #174 PR but this implementation is less hacky

"OBEY" will now accept/swallow pill (with ~99.9% chance if other actions are present). i didn't want to mess with "priority" and SexEngine doesn't filter unlikely actions the way InteractionSystem does, hence a silly number of 10000.0 score. feel free to adjust

without this change "OBEY"ing pc would pick unrelated actions like rubbing, ignoring the pill

"OBEY"ing pc would also refuse or spit pill out if they hate drugs:
\- when forcedObedience isn't high enough for pc to have all fetishes as Loves but "OBEY" appeared as result of RNG
\- or when being \*forced\* a drug, where spitting pill out has base score of 0.1

https://github.com/Alexofp/BDCC/blob/924c60c250b22c9cdba7884af8f3ad7f032a67a6/Game/SexEngine/SexActivity/DomDrugUse.gd#L403

don't know if i'll ever get used to reading those lines \~w\~

that one might need rewrite as resisting drug lover has -0.7 resist score but obedient drug lover has 0.1 resist score

\-

i'll try to be slower with pull requests now \~w\~ most of the things i wanted to help improve should be already part of the game, there might be another smol thing or two at some point

### DIFF
--- a/Game/SexEngine/SexActivity/DomDrugUse.gd
+++ b/Game/SexEngine/SexActivity/DomDrugUse.gd
@@ -384,32 +384,39 @@ func getSubActions():
 	var actions = []
 	if(state == "offering"):
 		var drugInfo = getDrugInfo(usedItemID, usedUniqueItemID)
-		var theScore = subInfo.personalityScore({PersonalityStat.Naive: 0.2, PersonalityStat.Subby: 0.2}) + subInfo.fetishScore({Fetish.DrugUse: 1.0})
+		var theObeyScore:float = subInfo.personalityScore({PersonalityStat.Naive: 0.2, PersonalityStat.Subby: 0.2}) + subInfo.fetishScore({Fetish.DrugUse: 1.0})
 		if(!getSub().isBlindfolded()):
-			theScore = (max(0.0, theScore) + drugInfo["scoreSubScore"]) * subInfo.getComplyScore()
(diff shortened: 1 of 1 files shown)


Alexofp:
My problem with this is that you might get into a situation where your character swallows a pill that you -really- didn't want. Someone might be into hypno but heavily against their character getting transformed. Like, hypnotized pc might eat a MorphIn pill and now they're screwed x3. I'm not sure what to do really x3

keerifox:
hmm could be a game setting or goal weight you could adjust?

to either have the pills apply no effect if consumed during OBEY, or don't offer/force the pills while pc is in OBEY (we simulate both dom & sub turns simultaneously while `subInfo.shouldFullyObey()` is still true, so it shouldn't be as difficult as it sounds)

personally i have forced TFs disabled in the settings so they shouldn't offer MorphIn

Alexofp:
I will make it so the UseTFDrug goal will never be generated against a hypnotized player that doesn't have the TF fetish

Alexofp:
And yeah, if you have the forced TFs turned on, all bets are off x3

Alexofp:
Should be okay after the tweak, thank you

PR #181, opened 2025-05-08
restraintkey loot list

keerifox:
should the loot table look closer to this, or is it intended that you can get up to 8 restraint keys if you simultaneously succeed the 75%, 50% and 10% RNG checks?

this is most noticeable when looting junkieStash and your character barely carries any keys, so sometimes you get A LOT of keys and i wonder if that's by design

### DIFF
--- a/Inventory/LootLists/RestraintKeyList.gd
+++ b/Inventory/LootLists/RestraintKeyList.gd
@@ -21,7 +21,7 @@ func getLoot(_id, _characterID, _battleName):
 		chanceMod *= 0.3
 	
 	return [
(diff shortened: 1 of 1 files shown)


Alexofp:
Yeah, my numbers look a bit silly x3

Alexofp:
Thank youuu

Alexofp:
Yeah I should have. I spent all my time with the pick up lines yesterday x3

Alexofp:
The first link actually only gives extra keys if you have smart-locked stuff on you. But the numbers are probably a bit too high anyway, I can poke the numbers

Issue #185, opened 2025-05-19
Some codes couples for `getName` or `getVisibleName`


CKRainbow:
Some codes directly or indirectly display the ID into the screen, which couple logical IDs with visible names. For example:

`getName` in

https://github.com/Alexofp/BDCC/blob/38ccd782d319326bba03b3d6a8965e154997ffad/Characters/Dynamic/CharacterType.gd#L1-L17

If I submit a PR trying to make the code more decoupling, would you accept the PR?

Alexofp:
It doesn't matter much I think. If you want

CKRainbow:
> It doesn't matter much I think. If you want

Ok, thanks. Actually it's for translation, it's better every text displayed is a literal string or something

PR #186, opened 2025-05-19
Refactor: Decoupling for some class

CKRainbow:
Do as #185 says.

### DIFF
--- a/Characters/Dynamic/CharacterPool.gd
+++ b/Characters/Dynamic/CharacterPool.gd
@@ -30,3 +30,20 @@ static func characterTypeToPool(charType):
 	if(charType == CharacterType.Engineer):
 		return Engineers
 	return null
+
+static func getVisibleName(poolID: String):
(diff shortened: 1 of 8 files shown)


Alexofp:
There was a mod that adds wings. It uses a unique bodypart slot name I think.. maybe it's worth to make an exception for the bodypart slot and return the id itself rather than error if nothing matches x3

CKRainbow:
> There was a mod that adds wings. It uses a unique bodypart slot name I think.. maybe it's worth to make an exception for the bodypart slot and return the id itself rather than error if nothing matches x3

I'm not sure but is there a way to add a slot instead of directly covering the original BodypartSlot.gd file? If not, I think it won't affect the wing slot mod.

Alexofp:
Oh, I see. I never actually checked the mod to see how it does it x3. Then yeah, it should be fine, thank youuuuuuuu

PR #187, opened 2025-05-24
Fix present-flirt failing too often, add text lines which may reveal interests

keerifox:
flirt by presenting body/bodypart practically never worked for me (which i think happens from negative values, abs of `Interest.Hates` would add 2 to divisor, abs of `Interest.Dislikes` would add 1 to divisor, even if player didnt have that trait they'd end up at 0 out of 2, at best)

i also wanted it to actually tell why it didn't succeed, at least sometimes. it reveals preferences in the process, but i think's fair since failed flirts make lust/affection go down by quite a bit

overall it looks like this

An androgynous body really drew their attention, though they're still seemingly unhappy about something. (57.8% success rate)

You were able to notice their aversion to any used hole. (58.7% success rate). You learned something new about their likes and dislikes..

### DIFF
--- a/Game/InteractionSystem/CharacterPawn.gd
+++ b/Game/InteractionSystem/CharacterPawn.gd
@@ -530,13 +530,31 @@ func getHowMuchLikesPawn(otherPawn, isClamped:bool = false) -> float:
 	
 	return lust.getOverallLikeness(otherPawn.getChar(), isClamped)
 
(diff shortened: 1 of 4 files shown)


Alexofp:
How the heck does one make a fun but simple flirting/talking minigame x3. Yoloooo thank youu

Issue #202, opened 2025-07-20
Description choosing mechanism in NPCSlave.gd not working.


12hydrogen:
Current version: 0.1.9fix1

There exists a method called "getPerfectIdleMessage()" in NPCSlave.md which should return the most proper description for current subject based on status of such subject. It contains some "distance" var which should represent how far this status is from that of a certain message, but the code seems doing the opposite:

```python
for dbEntry in db:
	var mainDistance = 0.0
	var supportDistance = 0.0
	var missingStatsDistance = 0.0
	
	var foundMainStats = 0
	var foundSupportStats = 0
	var theStats = dbEntry["stats"]
	for stat in theStats:
		if(mainStats.has(stat)):
			var distance = abs(mainStats[stat] - theStats[stat])
			mainDistance += 1.0 - distance
			foundMainStats += 1
		elif(supportStats.has(stat)):
			var distance = abs(supportStats[stat] - theStats[stat])
			supportDistance += 1.0 - distance
			foundSupportStats += 1
	
	missingStatsDistance = Util.maxi(0, mainStats.size() - foundMainStats) * 2.0 + Util.maxi(0, supportStats.size() - foundSupportStats) * 1.0
	
	var finalScore = mainDistance*2.0/(foundMainStats+1) + supportDistance*1.0/(foundSupportStats+1) - missingStatsDistance*0.1
	
	if(finalScore > highestScore):
		highestScore = finalScore
		bestMessage = dbEntry["message"]
```

Here, the distances are increased by `1.0 - distance`, and this `distance` is the absolute difference between desired value and actual value.

As a result, this method returns the most improper message rather than the most proper one. I've built a copy of this logic to check the cause of incorrect return value, and the result suggests that the `1.0 - distance` should be corrected to some other expressions, such as a simple `distance`.

This mechanism was created by others, so I do not fully understand their original idea, which may lead to an improper fix if I try to fix it myself, and I think this problem should be fixed by someone who created this logic. @Alexofp

12hydrogen:
A further explanation: In score calculation, when a status is missing in `dbEntry`, its contribution to `missingStatsDistance` exceeds the contribution of an existing status to `mainDistance` or `supportDistance`. Thus, the distance var should be a positive number to revert the influence of those missing statuses. However, the whole logic might be messed up when fewer statuses are missing in `dbEntry`. Currently this simple change should be doing well, returning a more proper message than former one.

PR #203, opened 2025-07-20
Possible fix for logic mistake

12hydrogen:
#202 A possible fix for problem mentioned in this issue.


Alexofp:
I didn't forget or purposefully ignored you, sorry x3. I'm just very busy with other stuff, I would have come around eventually. Sorryyy

PR #209, opened 2025-08-07
Adds an input field for searching datapacks and mods

PosionFox:
it is case insensitive and supports author search by prefixing with @

https://github.com/user-attachments/assets/77a44b40-8233-4b66-bea4-300f28284911


Alexofp:
That's very nice, thank you

PR #210, opened 2025-08-09
Fix able to pee into containers while portal panties is on.

MarsDDDDD:
Created two new buffs specifically for this one check.

### DIFF
--- a/Game/BaseCharacter.gd
+++ b/Game/BaseCharacter.gd
@@ -1380,6 +1380,16 @@ func hasReachableAnus():
 		return false
 	return hasAnus()
 
+func hasReachablePenisForUrination():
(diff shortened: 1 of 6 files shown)


Alexofp:
I think this might be an overkill solution. The problem is that the portal panties will have that 'blocks urination' buff visible even though its only for one small mechanic.
I would honestly just add a `isWearingPortalPanties()` check into the peeing scene, seems like a much simpler solution

MarsDDDDD:
> I think this might be an overkill solution.

You're right, I did it with scalability in mind because I don't know if either peeing or portal panties mechanics will get expanded upon e.g. more occasion to pee, more bathroom control contents or more interaction between portal panties and other stuffs in the future. I'll be making another request.

PR #212, opened 2025-08-10
added name generators

PosionFox:
Added two markov chain generators for npc names, it feeds off the names already present from RNGData

some improvemente like adding n-gram and max length options could be made

### DIFF
--- a/Game/MainScene.gd
+++ b/Game/MainScene.gd
@@ -40,6 +40,8 @@ var dynamicCharactersPools = {}
 var loadedDatapacks = {}
 var datapackCharacters = {}
 
+var npcNameGenerators = {}
+
 signal time_passed(_secondsPassed)
 signal saveLoadingFinished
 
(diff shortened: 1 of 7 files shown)


PosionFox:
note "1111" is my test character

PosionFox:
"What happens next depends only on the state of affairs now."
Markov chain is a process that makes events happen depending on previous events, instead of picking at random, it will decide based on the information you give to it, you can basically use it to generate more varied names that still resemble human speech and you can tweak the randomness by changing the n-gram between 1 and 4. Since it needs already defined names to construct its data model, you could feed it more fantasy-like names like the ones from TES (sheogorath, hasphat antabolis, etc) and it will still work, it is up to ones preference.

PosionFox:
This is an image with some generated names with order 3:

and the point is not to replace the list of names because it wouldn't work without it, it's just an option to add more variety, i could also remove the hardcoded bit in mainscene and let modders add their own more radical generators like a dwarf name generator or something but it's really up to you

PosionFox:
that is a fair point, personally i'm a fan of probability and procedural generation, i love games like tboi, soulash 2, dwarf fortress, caves of qud, valheim, etc. Some games lean more towards rng, others towards tailor made, others both, it is up to Rahi to decide what approach to take when adding new content.

PR #219, opened 2025-09-08
Toogleable alternative default save file naming scheme

shadowlin21:
A minuscule QoL.

Adds a toogleable option under "**Options/Saves**" which changes the default save naming scheme from: "_**{PC Name} - {Location}**_" to "_**{PC Name} - Save {NumOfSaves + 1}**_"

inb4, yes, I did make it mainly for myself because I prefer to have numbered saves for easier identification outside the game and it's a hassle renaming it every time, especially that my touch keyboard doesn't show up inside Godot text inputs. o7

### DIFF
--- a/Game/Options/GlobalOptions.gd
+++ b/Game/Options/GlobalOptions.gd
@@ -74,6 +74,8 @@ var cumIntensityMult = 1.0
 var visibleWritings:bool = true
 
 var autosaveEnabled = true
+var incrementalSaveNames: bool = false
+var saveNameCounter: int = 1
 
 var genderNamesOverrides = {}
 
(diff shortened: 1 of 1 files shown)


Alexofp:
I think it might be better to avoid adding a new option in this case. Might as well just combine the schemes. Something like:
{PC NAME} - Save {X} - {Location}
I'd store the current save number in the save itself too maybe. That way separate characters would have their own counters

Alexofp:
I added it in https://github.com/Alexofp/BDCC/commit/49d2dc93692462b6107453b1922b1481d949b1be Basically the game will now offer you this save name: <character_name> - Save X - <location_name>
The number X is stored in the save itself. So if you start a new game, the game will start counting from Save 1 again
Should be good enough probably

Alexofp:
Thank you, no reason to keep this open now I think

PR #224, opened 2025-09-21
Fix a temporary severe error when loading a save

CKRainbow:
After the recent update, Sometimes when you load a save, the game will crash or just show the start scene. like below (in Chinese sry)

QQ20250922-002152
QQ20250922-002206

This will happen because when you load a save, the game needs to change to the MainScene, while godot will only load the Node at the next idle frame, so `yield(get_tree(),"idle_frame")` is nessacery. But in the next idle frame, the node loading processes goes with the function `loadGame`. If you are not lucky enough, the `loadGame` will hit `GM.main` first, before the `MainScene` node is fully loaded.

So I used `call_deferred` to trigger the `loadGame` at the end of the frame, which may fix the problem.

For other solution, like using Signal, you can try it if you like.

### DIFF
--- a/Game/SaveManager.gd
+++ b/Game/SaveManager.gd
@@ -164,7 +164,7 @@ func loadGame(_path):
 func switchToGameAndLoad(_path):
 	var _ok = get_tree().change_scene("res://Game/MainScene.tscn")
 	yield(get_tree(),"idle_frame")
-	loadGame(_path)
+	call_deferred("loadGame", _path)
 
 func switchToGameAndResumeLatestSave():
 	var saves: Array = getSavesSortedByDate()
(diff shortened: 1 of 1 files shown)


Alexofp:
Huh, never happened to me, weird

Fox2Code:
I think having two `yield(get_tree(),"idle_frame")` instead of one would be equivalent to the `call_deferred`, this is what BDCC does for loading in `GlobalRegistry.gd`.

Alexofp:
I do 2 `yield(get_tree(),"idle_frame")` in global registry because opengl graphics are double-buffered. You need to swap framesbuffers 2 times to see stuff change. This is something different

CKRainbow:
[test_failload.zip](https://github.com/user-attachments/files/22455055/test_failload.zip)
I can always reproduce it when using web version. This is a simple save which created at web, and just start the game, doing nothing. When you load the save, there is chance the web freezes, especially loading it from the main manu instead of in-game.

Alexofp:
The chinese version looks pretty cool btw, good job ^^

PR #228, opened 2025-10-25
Sex work dom/sub requests, more teasing flavor text, restraints check

keerifox:
## dom/sub requests

allows to ask sex worker for the opposite role (dom/sub), instead of re-approaching them ~1-5 times in hopes they have a different option to offer

Option called 'I need a dom'

Client asking for sub to be a dom instead

\-

pawns may also do this to the player character

Client rudely asking for dom to be a sub instead

The client is not interested in a Service Dom option

## teasing flavor text

added couple more lines for flavor. no longer winking at client when blindfolded

Sex worker playfully sways hips

## restraints check

no longer able to offer "Usual" or "Pricy slut" options to pc/pawn clients in restraints that would normally prevent them from being a dom in most other interactions

## notes

feel free to change whatever, as usual i try to keep things clean and sensible but im only fox so

### DIFF
--- a/Game/InteractionSystem/Interactions/Prostitution.gd
+++ b/Game/InteractionSystem/Interactions/Prostitution.gd
@@ -413,7 +413,7 @@ func slut_won_do(_id:String, _args:Dictionary, _context:Dictionary):
 
 
 func slut_won_kick_text():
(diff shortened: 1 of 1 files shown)


Alexofp:
Seems nice, thank you. I will go over this when I have time, meow..

PR #236, opened 2025-11-03
add NpcCon.NotInStocksOrEquivalent to hypno encounter

### LINKED ISSUE #154: [0.1.7f1] NPC stuck in Stocks may trigger hypnosis event elsewhere
Steps to reproduce:
1. activate skill Hypnosis|Famously Suggestible
2. remove most Inmate NPCs
Note: In my cases, the NPC's preference for 'Hypnosis (hypnotist)' was neutral or above.
3. lock all other Inmate NPCs into the Stocks
4. walk around aimlessly and wait to be hypnotised

=> Eventually an NPC will attempt to hypnotise the PC. If you check on them afterwards, they are still in the Stocks.

Expected behaviour: On-map NPCs should only trigger the interaction, if they are in the same tile. If they want to, they should have to path to the PC. If they _cannot_, well, tough luck.

keerifox:
this fixes #154 but.. i don't imagine `grabNpcIDFromPoolOrGenerate` gets used often with an expectation to get someone in stocks/slutwall, so it might make sense to have this condition on by default, perhaps with a way to disable it when you in fact do want critters in stocks/slutwall to sometimes be part of the results

feel free to do it differently, it didn't particularly take me long

### DIFF
--- a/Characters/Dynamic/NpcCon.gd
+++ b/Characters/Dynamic/NpcCon.gd
@@ -24,4 +24,5 @@ enum {
 	HasReachablePenis,
 	HasReachableAnus,
 	IgnoreDisabledEncountersSetting,
+	NotInStocksOrEquivalent,
 }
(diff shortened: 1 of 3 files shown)


Alexofp:
I can code that when I'm not lazy

Alexofp:
Should be fixed by https://github.com/Alexofp/BDCC/commit/5d8fba3af1d7f36055efd3ba2c5b4e7834181af8
Npcs who are in stocks/slutwall have an interaction running so they will count as 'busy'

keerifox:
@Alexofp seems gooood. feel free to close #154 too

PR #239, opened 2025-11-13
Add "become" console command + make help reference aware

Fox2Code:
The "become" command allow the player to become any character id, only swapping body parts, skin, and visible elements.

Does not copy items, inventory, or equipment.


Alexofp:
Wouldn't a debug action be better

Fox2Code:
@Alexofp I use my last 2 working brain cells to make this, how much I can code depends on my health, I'll see what I can do about it when I am able too, but it might take a while, not been feeling very well recently.

PR #241, opened 2025-11-19
consistency stuff

Selinoccino:
- captain wright now acknowledges your victory or defeat after fighting risha
- dr quinn will take you to the proper cellblock, not the orange one every time
- your cell is mentioned as having a stool in the intro and a chair upon entering; changed that
- captain wright uses the verb "needs" despite pronouns, which causes a slight issue with they/them pronouns; changed to use youVerb instead
- also "an inmate number x" changed to "inmate number x", because otherwise i think it's implied that there are multiple inmates with that number
- dr quinn giving the pc water with the spacecillin is mentioned but the pc drinking it is not; fixed that
- corrected a few spelling and grammar errors
- simplified pc's line about risha after waking up the first time
- ejaculating when masturbating from the "me" menu thing now properly mentions and handles condoms (with breaking!), penis pumps (w/ collecting), and chastity cages (some text variation)

### DIFF
--- a/Game/LustCombat/LustActions/CumCum.gd
+++ b/Game/LustCombat/LustActions/CumCum.gd
@@ -31,11 +31,30 @@ func doAction(_lustState: LustCombatState, _args):
 		text += "\n\n"
 	
 	if(pc.hasPenis()):
+		var ejacPlace : String = "all over the floor"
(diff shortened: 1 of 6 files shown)


Alexofp:
Seems good, thank youuuuuuu

PR #244, opened 2025-11-19
pronoun selection tweaks

keerifox:
when asking player for the character's pronouns, show the pronouns themselves instead of genders

i've tried to adjust the wording to make less assumptions, since genders aren't always associated with specific pronouns, they/their isn't necessarily androgynous, non-gendered creatures aren't always it/its, and auto-selected pronouns are just something we assume/derive when trying to correctly refer to someone

\-

also shows current gender in a pretty color, since there is a button "keep your current gender" but you might forget what it currently is.. or just as a preview for chat color

Pick your character's gender. This will affect the color of your speech and how others treat you. Current gender: Androgynous, with the word androgynous colored in purple.

### DIFF
--- a/Scenes/CharacterCreatorScene.gd
+++ b/Scenes/CharacterCreatorScene.gd
@@ -29,11 +29,12 @@ func _run():
 
 	if(state == "pickpronouns"):
 		say("Pick your character's pronouns. This can be changed at any point")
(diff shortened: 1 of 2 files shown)


Alexofp:
Thank youuuuuuuuuu

PR #245, opened 2025-11-22
minor changes, misgendering & verb fixes, punctuation & spelling changes

Selinoccino:
- changed the phrase "digital century" to "digital age", as it's mentioned that the ai-operated mindtest has been in use for hundreds of years
- fixed a bit of misgendering, and fixed the verbs used in those cases as well
- changed all of the ellipses to have three dots (hopefully)
- changed all of the single quotation marks to a different character i found in the introScene file (which i'm guessing was used to avoid some issue when formatting the text, at least initially?)
- fixed some grammar and spelling mistakes
- cleaned up some small things (i think maybe only one)
- added some flavor text to the nurseryTalkScene
- added "themself" and "itself" in the array that calls himselfHerself func in the gameParser script


Selinoccino:
> I use 2 dots everywhere, don't change that x3

alright, i changed them back (w/ some more changes)
i also noticed that some ellipses in some scenes do have three dots instead of two, is there a general rule of thumb when to use which one? /genq

Alexofp:
That's a lot of fixes :o all seems good at the first glance. Huge thank youuuuu

PR #248, opened 2025-11-24
computer rewrite

Selinoccino:
- rewrote the entirety of the hacking minigames thing
- you caan now have multiple computers ("servers") with their own ip and commands, as well as admin commands, private files, and other stuff
- some commands are now integrated into the minigame itself and are available to all servers by default
- you can create your own commands for special servers, such as the open command for the safe
- this whole system should be more easily expandable, and more readable maybe

should be functionally about the same, judging by a fair bit of testing i did
i also tried to write comments for documentation purposes

### DIFF
--- a/Game/Computer/NetworkedComputerBase.gd
+++ b/Game/Computer/NetworkedComputerBase.gd
@@ -1,0 +1,275 @@
+extends ComputerBase
+class_name NetworkedComputerBase
+
+var connectedTo = ""
+
+var nextFreeId = 0
+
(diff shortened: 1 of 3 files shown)


Alexofp:
The result does look pretty clean. And it's good that you made a special class for it so you can still make computers in the old style. I will leave copies of the old Tavi computers, just as reference

But if someone reports bugs with this system, it's on you x3. Just being silly

Thank youuuuuu

PR #251, opened 2025-11-29
toggleable restrictions for tallymarks and bodywritings, begging for bodywritings

keerifox:
was requested in the itch.io comments

my understanding of this part of the codebase is a bit limited, so it's possible some tweaks would be needed, but in my limited testing everything seemed to work the way i expected


Alexofp:
I'd maybe replace 'Draw on body (Text)' with 'Add bodywritings'

PR #258, opened 2025-12-09
search & mod order drag n drop

Selinoccino:
- added a search function for the mod launcher, which shows mods with a similar name to the search query, as well as the two neighboring mods of the selected mod
- mods now show their position in the mod order, useful?
- you can now drag and drop mods around your mod order, and the order is saved when you release the mouse button
- also the mod order now shows how many mods you have next to the search bar thing

(...it was indeed annoying :/)


Alexofp:
You're insane x3 don't feel like you have to do stuff that I say
I will check when I can, meow meow

Alexofp:
Big thankiessss, I might touch some code, hope its okay

PR #259, opened 2025-12-10
Fix shell_open on Mac OSX platform

Meepyneepy:
For as long as I could remember, any button that opens a directory in File Explorer/Finder would give "Error -50" when trying to execute "OS.shell_open". So I created this to fix it. It's worked for me so far without issues.

I just replaced all instances of "OS.shell_open" with my custom shell_open function located in Util.


Alexofp:
Thankiesss

PR #267, opened 2025-12-20
Use FluidProduction for pee rather than hardcoding fluid amount as 100-500ml

FrivolousBorks:
While making a mod, I wanted to actually track pee production but it turns out there was no mechanism for it. This corrects that by adding a new `FluidProduction` class dedicated to pee. This probably needed to happen anyway because a lot of stuff involving watersports will need to track pee, so it kinda works out.

It is unusual for `FluidProduction`s in that it is not tied to a particular bodypart. This is largely a practical decision: bodyparts can currently only have the `FluidProduction` and you can have both sets of genitals, which would make it confusing. This could cause problems if code that invokes pissing makes certain assumptions, but I find it unlikely because people will likely just use the helper functions from `BaseCharacter`.


Alexofp:
I'd probably use a more neutral name such as urine instead of pee but it's fine x3. Thankiess

PR #268, opened 2025-12-29
character creator: allow quick switching between bodyparts, try not to reset per-bodypart colors/skin

keerifox:
like #156 but for bodyparts (head, hair, tail etc.)

if the player switches between bodyparts /after/ changing the colors/skin of said bodypart, game will try to carry those colors/skin over to the new bodypart

selecting "Nothing" bodypart on Tail slots etc. still deletes tail colors/skin as would have to save/load those values otherwise, feel free to do it if it seems worth..


Alexofp:
Seems reasonable, thank youuuuuuu

PR #269, opened 2025-12-31
slip by scene: fix animation showing guard facing the player when they're looking away

keerifox:
text: her back is turned away

guard: \*stares at you intently\*


Alexofp:
I'm not sure if these are even used anymore x3. Thankiesss

keerifox:
they're in umm Tavi's Quest2 when making way through the command deck towards captain's office

Alexofp:
Right, makes sense x3

PR #271, opened 2026-01-06
Send SexEvent when peeing inside someone

FrivolousBorks:
Realized when writing something that it was hard to have stuff happen when peeing occurred during sex. No way to have that happen right now obviously but we plan ahead. :3


Alexofp:
Thankiess

PR #274, opened 2026-01-21
Add support for SexReactionHandlers in modules

FrivolousBorks:
Made a mod and realized modules have to manually invoke GlobalRegistry for these. Seems pointless. Me fix.


Alexofp:
Thankiesss

PR #275, opened 2026-01-25
Make room transitions form after WorldEdits are applied

FrivolousBorks:
Currently when you add rooms with a `WorldEdit`, it doesn't add transition lines. You can fix this manually but it kind of sucks that you have to. It can be better.

This changes the game so that world edits are applied and _then_ transitions are added. Now world edit rooms have lines.

Bonus: While I was exploring this, I noticed that transitions were added twice, and also the filter for flood IDs wasn't working so the game was adding transitions for the entire game every time a drug den floor was generated. I fixed this. Now lines should only be added once.

Won't help with dynamically added rooms but there's only so much you can do.


Alexofp:
Good catch, thankiess

PR #277, opened 2026-01-29
Preserve camera and location when in rooms added by WorldEdits

FrivolousBorks:
When a room is added with a World Edit, the player's location is checked before the World Edit is applied so they're always reset to their cell.

Similarly, if they are in a Scene, this happens with the last camera location. You end up with some funky stuff if you load into a room added by a World Edit inside a Scene:
image
image

This fixes this by just moving the check for PC location and setting the camera to after World Edits are applied.


FrivolousBorks:
Hmm. This actually causes some saves to be softlocked. I am going to investigate further.

PR #278, opened 2026-01-30
Make it so location saves correctly for rooms added by World Edits

FrivolousBorks:
Let's try this one more time.

I noticed that saving in a room that was added by a World Edit always caused the player's location to get reset. Separately, I noticed that saving during a scene that happened to take place in an added room caused the camera to have a moment.

I have fixed these problems. The player's location is now checked after World Edits are applied, and the camera aim is saved at the start of scenes.

I also fixed a softlock that happens when loading a save that's in the middle of a scene that doesn't exist. It made testing this annoying.

image
image

### DIFF
--- a/Game/MainScene.gd
+++ b/Game/MainScene.gd
@@ -283,6 +283,7 @@ func getNewUniqueSceneID(blockedIDS=[]) -> int:
 	return result
 
 func runScene(id, _args = [], parentSceneUniqueID = -1):
+	GM.world.set("lastAimedRoomID", GM.pc.getLocation())
 	var scene = GlobalRegistry.createScene(id)
 	assert(scene != null, "SCENE WITH ID "+str(id)+" IS NOT FOUND. MAKE SURE IT WAS REGISTERED INSIDE THE MODULE.")
 	scene.uniqueSceneID = getNewUniqueSceneID([parentSceneUniqueID])
@@ -606,6 +607,9 @@ func loadData(data):
 	applyAllWorldEdits()
 	GM.world.addTransitions()
 	GM.pc.checkLocation()
+	var lastAimedLocation = GM.world.get("lastAimedRoomID")
+	if(lastAimedLocation != null && lastAimedLocation != ""):
+		GM.world.aimCamera(lastAimedLocation, true)
 
 func saveCharactersData():
 	var data = {}
--- a/Game/MainScene.gd
+++ b/Game/MainScene.gd
@@ -565,6 +565,9 @@ func loadData(data):
 		var id = SAVE.loadVar(sceneData, "id", "error")
 		
 		var scene = GlobalRegistry.createScene(id)
+		if(scene == null):
+			# If a scene doesn't exist, abort so the game is at least playable
+			break
 		add_child(scene)
 		sceneStack.append(scene)
 		print("Starting scene "+id)
--- a/Game/MainScene.gd
+++ b/Game/MainScene.gd
@@ -420,9 +420,6 @@ func loadingSavefileFinished():
 	#	GM.ui.getStage3d().resetToNothing()
 	reRun()
 	
-	applyAllWorldEdits()
-	GM.world.addTransitions()
-	
 	if(!rollbacker.rollbacking):
 		WHS.clearHistory()
 	IS.updatePCLocation()
@@ -606,6 +603,8 @@ func loadData(data):
 	#GM.world.updatePawns(IS)
 	#GM.world.setPawnsShowed(canShowPawns())
 
+	applyAllWorldEdits()
+	GM.world.addTransitions()
 	GM.pc.checkLocation()
 
 func saveCharactersData():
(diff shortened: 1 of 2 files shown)


Game/MainScene.gd
```diff
@@ -283,6 +283,7 @@ func getNewUniqueSceneID(blockedIDS=[]) -> int:
 	return result
 
 func runScene(id, _args = [], parentSceneUniqueID = -1):
+	GM.world.set("lastAimedRoomID", GM.pc.getLocation())
```
Alexofp: Why not?
`GM.world.lastAimedRoomID = GM.pc.getLocation()`

Game/MainScene.gd
```diff
@@ -565,6 +563,9 @@ func loadData(data):
 		var id = SAVE.loadVar(sceneData, "id", "error")
 		
 		var scene = GlobalRegistry.createScene(id)
+		if(scene == null):
+			# If a scene doesn't exist, abort so the game is at least playable
```
Alexofp: You can try to convince me but I dunno if this is needed. I didn't add any safeguards here on purpose so that the game crashes in an obvious way. If the scene just gets skipped, the game might end up in a bad (soft-locked potentially) state. Crashing loudly is better than doing the wrong thing silently..

Game/MainScene.gd
```diff
@@ -603,7 +604,12 @@ func loadData(data):
 	#GM.world.updatePawns(IS)
 	#GM.world.setPawnsShowed(canShowPawns())
 
+	applyAllWorldEdits()
+	GM.world.addTransitions()
 	GM.pc.checkLocation()
+	var lastAimedLocation = GM.world.get("lastAimedRoomID")
```
Alexofp: `var lastAimedLocation = GM.world.lastAimedRoomID`

Game/MainScene.gd
```diff
@@ -603,7 +604,12 @@ func loadData(data):
...
+	if(lastAimedLocation != null && lastAimedLocation != ""):
+		GM.world.aimCamera(lastAimedLocation, true)
```
Alexofp, line 609: I thought this is called already somewhere x3

Game/MainScene.gd
FrivolousBorks: Won't lie, I forgot that you can do that. The languages I work with daily don't allow you to index properties from other classes like that so it's easy to forget when you can. I will fix.

Game/MainScene.gd
FrivolousBorks: This is fair. I was annoyed by it when testing this fix but thinking about it more, I'm not sure that it will ever come up outside of developing mods. Anytime it does it's probably a mistake anyway, and catching it is good.

It was just annoying lol...

Game/MainScene.gd
FrivolousBorks: It's called when the `World` is loaded earlier. This happens before world edits are applied, so it needs to be called after the loading though. Could just move the loading down, but that causes problems because world edits need the world loaded to work.

After giving it a sniff test it seems like it's fine to remove it from World's `loadData` function as long as it's called here. Not sure why I didn't initially.

Game/MainScene.gd
```diff
@@ -283,6 +283,7 @@ func getNewUniqueSceneID(blockedIDS=[]) -> int:
 	return result
 
 func runScene(id, _args = [], parentSceneUniqueID = -1):
+	GM.world.lastAimedRoomID = GM.pc.getLocation()
```
Alexofp, line 286: This variable already gets set in aimCamera(). Do we need this call here?

FrivolousBorks: Normally it's unneeded because the camera is aimed at the player's location by the `WorldScene` when it runs. In cases where another scene is running though, the camera is never set, which leads to the fun camera state in the pictures above. To fix it, the camera has to be aimed _somewhere_.

The obvious solution is to have the variable default to the PC's location when the world loads, because that would mean that the existing `aimCamera` call in its `loadData` would sort it out. Unfortunately that doesn't work because of world edits though. The world needs to exist for world edits to be applied, but sometimes the room the PC is in won't exist until a world edit is applied. It's a miserable catch-22. :(

This leaves us with only a few solutions. I went with the decision to set the variable in `runScene` because the only alternative I can think of is putting `aimCamera(GM.pc.getLocation())` when scenes run. That has potential problems (if you run `FightScene` or `GenericSexScene` it'll reset the camera to the PC's location instead of where the scene takes place...) though, so I just didn't do it.

I'm open to alternative suggestions because I be but a humble dog, but I think this is probably the least invasive solution that doesn't create new problems for us. It makes sure the camera's intended location round-trips without relying upon the world being loaded, which is the important part.

Alexofp:
I will merge it but I will probably tinker a bit with the `GM.world.lastAimedRoomID = GM.pc.getLocation()` line, it's bugging me too much, sorry x3
Thank youuuuuuuu

Alexofp:
https://github.com/Alexofp/BDCC/commit/44d8e02ae632e0fea3e54e799ad8f2de5e3d4f7d
this should do the trick hopefully. The game will try to aim the camera at the stored room. If that fails, it will aim the camera at the player's location

PR #281, opened 2026-02-27
species code blocks

Selinoccino:
- added a code block that checks if an npc belongs to a species
- added a code block that checks if an npc is a hybrid

created a test scene in the datapack creator, tested them and they do seem to be working properly


Alexofp:
Seems good, thank youuuuuuuuuu

PR #282, opened 2026-02-27
no inappropriate item use in battle

Selinoccino:
- items that return false in canUseInCombat() func will not be usable in battle

i changed the permanent marker to be unusable in combat mostly for testing (and because someone said that it's supposed to be unusable in combat, but idk), i can remove that if needed

### DIFF
--- a/Scenes/InventoryScene.gd
+++ b/Scenes/InventoryScene.gd
@@ -227,6 +227,10 @@ func onInventoryItemSelected(item: ItemBase):
 		else:
 			addButton("Put on", item.getVisisbleDescription(), "puton", [item.getUniqueID()])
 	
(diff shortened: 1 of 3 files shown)


Scenes/InventoryScene.gd
```diff
@@ -227,6 +227,10 @@ func onInventoryItemSelected(item: ItemBase):
 		else:
 			addButton("Put on", item.getVisisbleDescription(), "puton", [item.getUniqueID()])
 	
+	if item.canUseInCombat()==false:
```
Alexofp: `if(!item.canUseInCombat()):`

You probably wanted that x3

Alexofp:
Thank youuuuuuuuuu. Yeah somehow I missed that

PR #284, opened 2026-02-28
networked computer improvements

Selinoccino:
- files are now sorted based on id, making it possible to use binary search to find files
- auto ids are now consistent and work properly
- binary search is now used to find files
- kinda fixed saving a bit
- other minor changes

the file id changes were a long time coming (since the first rework), so it's nice to finally get it out of the way

most of the weirdly complicated stuff will never need to be touched again (ComputerFileSorting and anything that uses bsearch), save for changing how files are stored/sorted (and in that case i could still work on fixing it as well)

i'll probably work on making the code more readable at some later point, because it's an eyesore right now

### DIFF
--- a/Game/Computer/NetworkedComputerBase.gd
+++ b/Game/Computer/NetworkedComputerBase.gd
@@ -1,32 +1,30 @@
 extends ComputerBase
 class_name NetworkedComputerBase
 
-var connectedTo = ""
-
-var nextFreeId = 0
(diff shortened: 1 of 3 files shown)


Alexofp:
I will trust you that it works x3. There is really no need to put so much effort into this terminal simulator thingie x3 But I know how it feels because I'm the one who decided to make it in the first place x3

Thank youuuu

PR #285, opened 2026-02-28
qol changes (mostly)

Selinoccino:
- made mainScene not check for every debug action even if it's already found one match
- you can now specify a sceneTag and arguments when running a scene from the debug menu
- the console no longer opens whenever you press the tilde key while typing into a control node (you can now write stuff like "i need him so bad~" in datapacks)
- console no longer uses the physical key code thing, and instead uses godot's input action mapping
- also made a hacky-ish way to still be able to toggle the console with the tilde key

seems to be working fine, and there shouldn't be much potential for errors/crashing/whatnot, but i didn't test very extensively

### DIFF
--- a/Game/MainScene.gd
+++ b/Game/MainScene.gd
@@ -282,11 +282,12 @@ func getNewUniqueSceneID(blockedIDS=[]) -> int:
 		result += 1
 	return result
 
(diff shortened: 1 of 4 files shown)


Alexofp:
Looks alright on the first glance x3 Thank youuuuuuuuuuu

PR #287, opened 2026-03-02
mod launcher improvements (?) & a few minor changes

Selinoccino:
- simplified CrotchBlockVisual gui input
- minor changes in ComputerSimScene to avoid redundant checks maybe
- launcher mod entries can now be toggled on and off by double clicking, for more intuitive control to accompany the new dragging mods around thing
- also removed the texturebutton from the launch mod entries to reduce node count, and replaced their functionality with a gui_input function instead, so it should be functionally the same
- also saved the launch mod entry's panel stylebox for use with creating drag previews, which should be much more consistent now. this will hopefully make it easier to sync style changes between the entry and its drag preview
- finally, made toggling mods its own function, and made it so that you can specify an entry other than the currently selected one (maybe useful for inverting mod selections altogether for some reason?)

### DIFF
--- a/Game/Datapacks/UI/CrotchCode/CrotchBlockVisual.gd
+++ b/Game/Datapacks/UI/CrotchCode/CrotchBlockVisual.gd
@@ -201,12 +201,8 @@ func _on_PopupMenu_index_pressed(index):
 	if(index == 0):
 		doSelfdelete()
 
-var lastRightClick = 0
 func _on_CrotchBlock_gui_input(event):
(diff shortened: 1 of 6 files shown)


Alexofp:
Node count isn't really a problem when we're talking about <1000 nodes anyway. And using an actual button instead of trying to emulate one is better I think because of stuff you'd never think about. Like switching focus with tab or something x3. Although in this case since its invisible it doesn't really matter

Seems good, thank youuuuuu

PR #288, opened 2026-03-04
fix combat items

Selinoccino:
i kinda messed up my last change, making all items that are unusable in combat always unusable. this fixes that

sorry

### DIFF
--- a/Scenes/InventoryScene.gd
+++ b/Scenes/InventoryScene.gd
@@ -227,7 +227,7 @@ func onInventoryItemSelected(item: ItemBase):
 		else:
 			addButton("Put on", item.getVisisbleDescription(), "puton", [item.getUniqueID()])
 	
-	if(!item.canUseInCombat() && fightMode):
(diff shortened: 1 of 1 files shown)


Scenes/InventoryScene.gd
```diff
@@ -227,7 +227,7 @@ func onInventoryItemSelected(item: ItemBase):
 		else:
 			addButton("Put on", item.getVisisbleDescription(), "puton", [item.getUniqueID()])
 	
-	if(!item.canUseInCombat()):
+	if(!item.canUseInCombat() && fightMode):
```
Fox2Code: I think `fightMode && !item.canUseInCombat()` would make more sense than `!item.canUseInCombat() && fightMode` because in the first expression the `canUseInCombat` call is skipped if `fightMode` is `false`.

Selinoccino: yeah that does make sense, made a commit changing it

Alexofp:
Thankiessss

PR #316, opened 2026-05-14
correctly handle singular they/them pronouns, typo fixes

keerifox:
prior to PR:

`{npc.You} {npc.youAre} stuck in the slutwall. {npc.YouHe} {npc.youAre} completely helpless. {npc2.You} {npc2.youVerb('walk')} past {npc.youHim}. {npc2.YouHe} {npc2.youVerb('want')} no part in this.`
> Jackie **are** stuck in the slutwall. They are completely helpless. Darlene **walk** past them. They want no part in this.

\-

after PR:

`{npc.You} {npc.youAre} stuck in the slutwall. {npc.YouHe} {npc.youAreHeIs} completely helpless. {npc2.You} {npc2.youVerb('walk')} past {npc.youHim}. {npc2.YouHe} {npc2.youHeVerb('want')} no part in this.`
> Jackie **is** stuck in the slutwall. They are completely helpless. Darlene **walks** past them. They want no part in this.

* * *

this was originally meant to be like a 2 line change, but {npc.You} produces a name while {npc.YouHe} produces a pronoun. both are frequently followed with {npc.youVerb}.. the name needs to have "is/does/has" after it, while the pronoun needs to have "are/do/have"


Alexofp:
That probably took forever, wow
Thank youuu
I wonder if there some cursed way to simplify it. Make the parser stateful. If we used {npc.youHe}, then the next {npc.youAre} becomes {npc.youAreHeIs} or whatever x3

PR #317, opened 2026-05-14
fix #313 Eliza Quinn interaction 'Milking' does not re-equip bra afterwards

### LINKED ISSUE #313: [0.2.5] Eliza Quinn interaction 'Milking' does not re-equip bra afterwards
It does re-equip panties though.

Expected behaviour: It should completely restore all previously equipped clothing.

Full list of affected interactions:
* Milking|Milk|Breast Pumps
* Milking|Milk|Milking machine
* Milking|Quick milking

(Everything that replaces the item in the 'breasts' slot. I see there has been a workaround for Chastity Cages: These just stay on, instead of being replaced with a pump.)

Zsar:
Implemented a quick-and-dirty fix on just the quick milking scene.

Found that fixing this properly would touch quite a lot of code, so I am asking for directions first.

... Seeing that one would have to retrofit any scene-by-scene with this capability makes me feel a bit nauseous. Even with a dedicated helper class, it would still mean two calls and one variable per scene, which may always be forgotten or otherwise go out of sync between code changes.

Maybe `SceneBase` could get its own inventory handling functions and encapsulate the helper class, such that individual scenes only need to decide whether items _should be_ replaced or not? That way only one call (per item) can be forgotten, the one that stores said item in the first place.

### DIFF
--- a/Inventory/Inventory.gd
+++ b/Inventory/Inventory.gd
@@ -340,8 +340,11 @@ func forceEquipRemoveOther(item) -> bool:
 	
 	return equipItem(item)
 
-func forceEquipRemoveOtherIfExists(item: ItemBase) -> bool:
-	return forceEquipRemoveOther(item) if item else false
+func forceEquipReturnOther(item: ItemBase):
+	var slot: String = item.getClothingSlot()
+	var displacedItem = removeItemFromSlot(slot)
+	equipItem(item)
+	return displacedItem
 
 func forceEquipStoreOther(item):
 	var slot:String = item.getClothingSlot()
--- a/Inventory/Inventory.gd
+++ b/Inventory/Inventory.gd
@@ -267,7 +267,7 @@ func getCharacter():
 		return get_parent()
 	return null
 		
-func equipItem(item):
+func equipItem(item) -> bool:
 	if(hasItem(item)):
 		removeItem(item)
 	
@@ -332,13 +332,16 @@ func unequipSlotRemoveIfRestraint(slot):
 		addItem(theitem)
 		return true
 
-func forceEquipRemoveOther(item):
+func forceEquipRemoveOther(item) -> bool:
 	var slot:String = item.getClothingSlot()
 	
 	if(hasSlotEquipped(slot)):
 		removeItemFromSlot(slot)
 	
 	return equipItem(item)
+
+func forceEquipRemoveOtherIfExists(item: ItemBase) -> bool:
+	return forceEquipRemoveOther(item) if item else false
 
 func forceEquipStoreOther(item):
 	var slot:String = item.getClothingSlot()
@@ -358,7 +361,17 @@ func forceEquipStoreOtherUnlessRestraint(item):
 			addItem(storedItem)
 	
 	return equipItem(item)
-	
+
+func forceEquipStoreOtherUnlessRestraintReturnOther(item: ItemBase):
+	var slot:String = item.getClothingSlot()
+	var storedItem = null
+	if(hasSlotEquipped(slot)):
+		storedItem = removeItemFromSlot(slot)
+		if(!storedItem.isRestraint() || storedItem.isImportant() || storedItem.isRestraintShouldKeep()):
+			addItem(storedItem)
+	equipItem(item)
+	return storedItem
+
 func equipItemBy(item, equipper):
 	var success = equipItem(item)
 	if(success):
--- a/Modules/MedicalModule/Milking/ElizaQuickMilkingScene.gd
+++ b/Modules/MedicalModule/Milking/ElizaQuickMilkingScene.gd
@@ -6,9 +6,7 @@ var vaginaMilked = false
 var hasPenisPump = false
 var amountCollected = 0.0
 
-var displacedItemBreasts: ItemBase = null
-var displacedItemPenis  : ItemBase = null
-var displacedItemVagina : ItemBase = null # Not currently in use because there is no pump item implemented yet. -Zsar 2026-05-14
+var displacedItems := LightInventory.new()
 
 func _init():
 	sceneID = "ElizaQuickMilkingScene"
@@ -23,13 +21,13 @@ func _reactInit():
 			var theFluids = thePump.getFluids()
 			if(theFluids):
 				theFluids.addFluid("Milk", 400.0)
-		displacedItemBreasts = GM.pc.getInventory().forceEquipStoreOtherUnlessRestraintReturnOther(thePump)
+		displacedItems.addItem(GM.pc.getInventory().forceEquipReturnOther(thePump))
 	if(GM.pc.hasReachablePenis() || GM.pc.isWearingChastityCage()):
 		amountCollected += GM.main.SCI.processMilkPlayerPenis()
 		penisMilked = true
 		if(GM.pc.hasReachablePenis() || GM.pc.getWornChastityCage().getRestraintData().canBeEasilyRemovedByDom()):
 			var thePump = GlobalRegistry.createItem("PenisPump")
-			displacedItemPenis = GM.pc.getInventory().forceEquipStoreOtherUnlessRestraintReturnOther(thePump)
+			displacedItems.addItem(GM.pc.getInventory().forceEquipReturnOther(thePump))
 			hasPenisPump = true
 	if(GM.pc.hasReachableVagina()):
 		amountCollected += GM.main.SCI.processMilkPlayerVagina()
@@ -61,9 +59,8 @@ func _react(_action: String, _args) -> void:
 			GM.pc.getInventory().clearSlot(InventorySlot.Penis)
 		if(breastsMilked):
 			GM.pc.getInventory().clearSlot(InventorySlot.UnderwearTop)
-		GM.pc.getInventory().forceEquipRemoveOtherIfExists(displacedItemBreasts)
-		GM.pc.getInventory().forceEquipRemoveOtherIfExists(displacedItemPenis)
-		GM.pc.getInventory().forceEquipRemoveOtherIfExists(displacedItemVagina)
+		for displacedItem in displacedItems.getAllItems():
+			GM.pc.getInventory().forceEquipRemoveOther(displacedItem)
 
 		playAnimation(StageScene.Duo, "stand", {npc="eliza"})
 		aimCameraAndSetLocName(GM.pc.getLocation())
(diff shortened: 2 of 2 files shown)


Modules/MedicalModule/Milking/ElizaQuickMilkingScene.gd
```diff
@@ -6,6 +6,10 @@ var vaginaMilked = false
 var hasPenisPump = false
 var amountCollected = 0.0
 
+var displacedItemBreasts = null
```
Zsar: Should probably become its own class 'ItemReEquipHelper' or similar, encapsulating re-equip behaviour in a single function.

```
displacedItems: ItemBase[]

func displace(item: ItemBase) -> void
func reEquipAll() -> void
```
or maybe
`func reEquipAll(discardDisplacers: bool) -> void`
or even
`func reEquipAll(displacersGoHere: Optional<Inventory>) -> void`
so one need not remove temporary items like the pumps manually and (in the latter case) may dynamically equip persistent items from an inventory without fear of accidentally destroying them.

Modules/MedicalModule/Milking/ElizaQuickMilkingScene.gd
```diff
@@ -19,14 +23,14 @@ func _reactInit():
...
-		GM.pc.getInventory().forceEquipStoreOtherUnlessRestraint(thePump)
+		displacedItemBreasts = GM.pc.getInventory().forceEquipStoreOtherUnlessRestraintReturnOther(thePump)
 	if(GM.pc.hasReachablePenis() || GM.pc.isWearingChastityCage()):
 		amountCollected += GM.main.SCI.processMilkPlayerPenis()
 		penisMilked = true
-	if(GM.pc.hasReachablePenis()):
-		var thePump = GlobalRegistry.createItem("PenisPump")
-		GM.pc.getInventory().forceEquipStoreOtherUnlessRestraint(thePump)
-		hasPenisPump = true
+		if(GM.pc.hasReachablePenis() || GM.pc.getWornChastityCage().getRestraintData().canBeEasilyRemovedByDom()):
```
Zsar, line 28: Can be dropped, if you like.

Seemed sensible, but is not in scope of the issue.

Modules/MedicalModule/Milking/ElizaQuickMilkingScene.gd
```diff
@@ -57,7 +61,10 @@ func _react(_action: String, _args):
 			GM.pc.getInventory().clearSlot(InventorySlot.Penis)
 		if(breastsMilked):
 			GM.pc.getInventory().clearSlot(InventorySlot.UnderwearTop)
-		
+		GM.pc.getInventory().forceEquipRemoveOtherIfExists(displacedItemBreasts)
```
Zsar: See [comment above](https://github.com/Alexofp/BDCC/pull/317#discussion_r3244654226): This could and probably should be a single function call, lest one forget some item.

Inventory/Inventory.gd
```diff
@@ -358,7 +364,17 @@ func forceEquipStoreOtherUnlessRestraint(item):
 			addItem(storedItem)
 	
 	return equipItem(item)
-	
+
+func forceEquipStoreOtherUnlessRestraintReturnOther(item: ItemBase):
```
Zsar, line 368: It would be better to change the return value of `forceEquipRemoveOther` and assorted functions to be a struct
```
successfully : bool
itemDisplaced: Optional<ItemBased>
```
but I did not want to make such a big change without asking you first.

(Don't mind the leg work, though, so if you give the go-ahead, I shall.)

... It also seems that GDScript flat out cannot handle null-ables as return types? So it would _also_ need such an Optional class, which it _also_ does not implement itself?
\- Damn, one should think that if someone were to create their own programming language, they'd take care to make it _better than its predecessors!_

Inventory/Inventory.gd
```diff
@@ -332,14 +332,20 @@ func unequipSlotRemoveIfRestraint(slot):
...
-func forceEquipRemoveOther(item):
+func forceEquipRemoveOther(item) -> bool:
 	var slot:String = item.getClothingSlot()
 	
 	if(hasSlotEquipped(slot)):
 		removeItemFromSlot(slot)
 	
 	return equipItem(item)
 
+func forceEquipRemoveOtherIfExists(item: ItemBase) -> bool:
```
Zsar: See [comment below](https://github.com/Alexofp/BDCC/pull/317#discussion_r3244654226): If a class were implemented, it would deal with this logic instead of this "free-standing" function.

Modules/MedicalModule/Milking/ElizaQuickMilkingScene.gd
```diff
@@ -6,6 +6,10 @@ var vaginaMilked = false
 var hasPenisPump = false
 var amountCollected = 0.0
 
+var displacedItemBreasts = null
+var displacedItemPenis   = null
+var displacedItemVagina  = null # Not currently in use because there is no pump item implemented yet. -Zsar 2026-05-14
```
Zsar: It is a bit sad, is it not? \*hint, hint, nudge, nudge*

Modules/MedicalModule/Milking/ElizaQuickMilkingScene.gd
Zsar, line 81: Tried adding the new variables here, but it seems that the loaded reference is invalid: It turns into `null` when converted to `ItemBase`.

... Every single step in the Godot debugger takes several minutes. How do you work with this!?

Modules/MedicalModule/Milking/ElizaQuickMilkingScene.gd
```diff
@@ -84,3 +94,9 @@ func loadData(data):
 	vaginaMilked = SAVE.loadVar(data, "vaginaMilked", false)
 	hasPenisPump = SAVE.loadVar(data, "hasPenisPump", false)
 	amountCollected = SAVE.loadVar(data, "amountCollected", 0.0)
+	var idDisplacedItemBreasts = SAVE.loadVar(data, "displacedItemBreasts")
+	displacedItemBreasts = GM.pc.getInventory().getItemByUniqueID(idDisplacedItemBreasts) if idDisplacedItemBreasts else null
+	var idDisplacedItemPenis = SAVE.loadVar(data, "displacedItemPenis")
+	displacedItemPenis   = GM.pc.getInventory().getItemByUniqueID(idDisplacedItemPenis)   if idDisplacedItemPenis   else null
```
Zsar: Mmh. Works for the underwear, does not work for the chastity cage. Investigating... ~~

Modules/MedicalModule/Milking/ElizaQuickMilkingScene.gd
Alexofp: This could error out if we're waring portal panties or something. The pp is not reachable but there is no chastity cage. I'd promoute it to a variable and check if its not null

Modules/MedicalModule/Milking/ElizaQuickMilkingScene.gd
Alexofp: Debugger works for me I dunno, sorry x3

Inventory/Inventory.gd
Alexofp: Do this really need extra new functions? Can you maybe use the other ones that exist already.
instead of doing `displacedItems.addItem(GM.pc.getInventory().forceEquipReturnOther(thePump))`
you can get the current item in that slot and unequip it manually 
```gdscript
var theItem = GM.pc.getInventory().removeItemFromSlot(thePump.getClothingSlot())
if(theItem):
   displacedItems.addItem(theItem)
GM.pc.getInventory().forceEquipStoreOtherUnlessRestraint(thePump)
```
Or something x3

Zsar:
Had a thought overnight: The ("non-soft") Player Slavery system already contains the functionality to store player equipment for arbitrary times and restore it afterwards, via `PlayerSlaveryHolder#transferAllItems`, `#storePlayersItems`, `#givePlayerItemsBack`.

It might be sensible to not reinvent the wheel here, especially seeing that #318 is basically the same issue as this one (just more destructive).

Should I "just" coopt the `PlayerSlaveryStash` and move `transferAllItems` et all into `SceneBase` or a similar central position?

Inventory/Inventory.gd
Zsar: I just wrote the extra function to have a working prototype, but I also dislike adding more. That's why I figured it would be more sensible to change the return value of the existing functions. Just did not want to dump 100+ code changes on you without asking first.

Of course, once there is a dedicated object to deal with these code changes, it does not harm to just do it in several lines - those would no longer pollute the Scene code, they would all be hidden behind a tidy one-liner calling on said object.

Candidates are
* LightInventory (which should then probably be renamed, but de-facto is already used exclusively for item transfers and temporary storage)
* PlayerSlaveryStash (which is a "character", I see, but thus could provide full slot information by equipping what was equipped and storing what was not)
* a new class (my thought yesterday, before discovering PlayerSlaveryStash; now I'd rather use one of the existing two)

Alexofp:
Using LightInventory is fine, I don't see a reason for anything more complicated
You can make it easier by 'unequpping items' instead (so they stay in player's inventory) and storing an array of unique IDs of those items and then re-equipping them after the scene end.

Zsar:
[That's what I had tried](https://github.com/Alexofp/BDCC/pull/317/changes/83382531f77c100089998cccdf8de10f8b83f288#diff-c920908e9664f36a6f30c67c8b23ae7713b57128666ff9d2d716da0097a8b95bL83-L102) before using a separate LightInventory, but it did not always work and I could not figure out why: For some reason, it would never re-equip specifically a Chastity Cage.

Debugging `loadData`, loading the uniqueId worked, but then`getItemByUniqueID` always yielded null, whereas a Laced Bra loaded just fine. I took care to try both having several of both items in inventory, as well as just a single one, and in all cases it was the Chastity Cage that failed to be retrieved.

Alexofp:
Pushed a version of a fix in https://github.com/Alexofp/BDCC/commit/c5ccd3fa2306bf83a9ef6b1dd91e8a0d410b5a63
I'm just unequipping the items temporarily (they get moved from the equipped items to the stored ones). I save the unique IDs of those items and then just re-equip them at the end. Should work well enough x3

I will close this. Thank youuuuuuuu

Zsar:
Ah, sorry for taking too long! Thank you.

Issue #311, opened 2026-05-02
[0.2.4] Alex Rynard may ask PC about his fatherhood of their pregnancy, before unlocking any sort of intimacy


Zsar:
(You can use the save from #310. Just quit the interaction and go start to him.)

I have not yet unlocked even the first Trust Exercise.

If memory serves, the last unlock was being allowed to fix Datapads on my own.
Indeed, the next 'Help' afterwards starts with
> You join Alex. And today is the first time you are allowed to help!

Expected behaviour: He should know about 'the birds and the bees'.

charredlizard:
Can confirm I was able to replicate this. It can happen in previous updates as well.

Broslan:
confirm

Zsar:
Mmh. The second check ought to deal with this:

```gdscript
if(!getModuleFlag("AcePregExpac", "Alex_ToldIsFather", 1)) || !GM.pc.isHeavilyPregnant():
	return false
if(!getFlag("AcePregExpac.Alex_CameInside", false)):
	return false
```

... _Should_ these flags be checked by different functions? They are both defined identically, so I'd expect them to also be checked identically:
```gdscript
func getFlags():
	return {
		"Alex_ToldIsFather": flag(FlagType.Number),
		"Alex_TimesMilkingPlayer": flag(FlagType.Number),
		"Alex_CameInside": flag(FlagType.Bool),
	}
```

Zsar:
Apparently this was stealth-fixed in 4a16ab2788bdfb4f2d5e40158df62a8cba0a5162 .

PR #320, opened 2026-05-16
#311 fix wrong flag scope and inverted default value

Zsar:
(Still need to test fix. Lost save for quick reproduction.)

### DIFF
--- a/Modules/AcePregExpac/AlexScenes/AlexToldHeIsFatherEvent.gd
+++ b/Modules/AcePregExpac/AlexScenes/AlexToldHeIsFatherEvent.gd
@@ -9,7 +9,7 @@ func registerTriggers(es):
 func react(_triggerID, _args):
 	if(!getModuleFlag("AcePregExpac", "Alex_ToldIsFather", 1)) || !GM.pc.isHeavilyPregnant():
 		return false
-	if(!getFlag("AcePregExpac.Alex_CameInside", false)):
+	if(!getModuleFlag("AcePregExpac", "Alex_CameInside", true)):
 		return false
 	
 	setModuleFlag("AcePregExpac", "Alex_ToldIsFather", 0)


Modules/AcePregExpac/AlexScenes/AlexToldHeIsFatherEvent.gd
```diff
@@ -9,7 +9,7 @@ func registerTriggers(es):
 func react(_triggerID, _args):
 	if(!getModuleFlag("AcePregExpac", "Alex_ToldIsFather", 1)) || !GM.pc.isHeavilyPregnant():
 		return false
-	if(!getFlag("AcePregExpac.Alex_CameInside", false)):
+	if(!getModuleFlag("AcePregExpac", "Alex_CameInside", true)):
```
Zsar, line 12: Double negation: was 'if not did not', should have just been 'if not'.

Also wrong flag scope apparently caused this wrong default value to always be used instead of the actual flag.

(At least that is my understanding from just reading the code. Will test under debugger to verify.)

Zsar: ... No, the more I look at it, the more I am sure, it _was_ fine. - Except for the "not working" part, that is!

Zsar: Added these
```gdscript
var _flagDefaultFalse = getFlag("AcePregExpac.Alex_CameInside", false)
var _flagDefaultTrue  = getFlag("AcePregExpac.Alex_CameInside", true)
var _moduleFlagDefaultFalse = getModuleFlag("AcePregExpac", "Alex_CameInside", false)
var _moduleFlagDefaultTrue  = getModuleFlag("AcePregExpac", "Alex_CameInside", true)
```
and lo: *both* `getFlag` and `getModuleFlag` always return the default value! ~~

... And here I thought this would be easy and straight-forward.

Zsar: Apparently the second check _is the fix_, and has been committed one day after the issue without updating it: 4a16ab2788bdfb4f2d5e40158df62a8cba0a5162.

Oh well, nvm. then!

Alexofp:
Yeah sorry I sometimes don't comment on issues after fixing them

PR #321, opened 2026-05-16
fix #303 NPCs locked in Stocks magically restore Stamina to 100% after 'Use'

### LINKED ISSUE #303: [0.2.3] NPCs locked in Stocks magically restore Stamina to 100% after 'Use'
Steps to reproduce:

1. find an NPC knocked down and in the process of being locked into the Stocks
(unsure whether doing it yourself or using the Slutwall instead makes a difference)
2. 'Use' the NPC once locked in the Stocks
=> 0 Stamina
3. take care to stop while they are still at 0
=> you may for a second see their Stamina bar refill
4. free the NPC with a Restraint Key
5. attack the NPC
=> 100% Stamina

Not 100% sure, but it seems this _does not_ happen, if an NPC 'Use's them.

Expected behaviour: They must 'Wait' or use perks to regain Stamina or let someone else with Stamina help them.

Zsar:
fixed for Stocks, TODO:
* ~~deduplicate (or at least fix separately) Slutwall~~ - actually works fine and now I cannot reproduce them ignoring Stamina when I go back to HEAD~1 either...
Guess I can drop 3bc120a263e9105c207f7cd48bb24bae8c65b1f1 then.
* further verify nothing else broke

### DIFF
--- a/Game/BaseCharacter.gd
+++ b/Game/BaseCharacter.gd
@@ -2305,11 +2305,6 @@ func afterSexEnded(sexInfo):
 		if(resultText != null && resultText != ""):
 			GM.main.addMessage(resultText)
 	
-	if(!isPlayer()):
-		addLust(-getLust())
-		addPain(-getPain())
-		addStamina(getMaxStamina())
-
 	consciousness = 1.0
 	arousal = 0.0
 		


Game/InteractionSystem/Interactions/InStocks.gd
```diff
@@ -20,17 +20,17 @@ func start(_pawns:Dictionary, _args:Dictionary):
 
 func init_text():
 	saynn("{inmate.You} {inmate.youAre} stuck in stocks, there is very little {inmate.youHe} can do, the movement of {inmate.yourHis} head and arms is blocked by a giant metal frame with 3 holes, its angle is forcing {inmate.youHim} to constantly stay bent forward, exposing {inmate.yourHis} butt, {inmate.yourHis} ankles are chained so {inmate.youHe} can’t really move them too. {inmate.YouHe} {inmate.youAreHeIs} completely helpless.")
-	var thePawn = getRolePawn("inmate")
-	if(!thePawn.isPlayer()):
-		saynn("{inmate.name} has "+str(getRoleChar("inmate").getStamina())+" stamina left.")
-	var npc = getRoleChar("inmate")
-	if(npc.hasEffect(StatusEffect.HasTallyMarks)):
-		var tallyMarks = npc.getEffect(StatusEffect.HasTallyMarks)
+	var inmate  := getCharByRole("inmate")
```
Zsar: I was not able to *spot* why NPCs did not regain Stamina or needed Stamina to struggle.

But luckily replacing all alternatives with `getCharByRole` fixed it.

Alexofp:
There are too many unrelated changes, why change the function definitions and stuff if you're just fixing one thing x3
I also dunno if all the other changes gonna break anything. I can only guess

Game/InteractionSystem/Interactions/InStocks.gd
Zsar: ... And now it does not happen anymore, even without the changes.

Alas.

Zsar:
Sorry, I guess.

It is inconceivable to me that anyone would want to memorise all this information, when they could simply write it down and have the tools help with not making mistakes, which were specifically designed to... help with not making mistakes.

(And of course, _I, who I am not familiar with it_ need to find this information from somewhere, and the IDE is no help: Rider only deduces these up to a few (2?) steps and then it gives up. The Godot editor does not even seem to try.)

But it's your project. Reverting.

Alexofp:
I'm sorry if I sound mean, didn't mean to. I'm tired, sorry x3
I appreciate the help, just don't have the capacity to think about this stuff currently. I could just fix it myself when I get into the mood

Zsar:
No harm done.

And please, do feel free to let me wait for as long as you want to or need to - I am only doing this right now, because I myself have the time right now.
The platform is asynchronous by design and it is good that way.

I do not mind digging into this as a way to get to grasps with Godot without having to roll my own project from scratch... but, well, Godot and particularly GDScript seem to be specifically designed to make contributing hard. (Reading and understanding the source, that is. Necessary precondition for _having_ anything to contribute.)

Maybe, once you have some time and mental energy for it, you could add a [CONTRIBUTING.md](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors#about-contributing-guidelines) to the repository re. type hints and similar coding guidelines. I saw that you had started - relatively late in development - to add and retrofit type hints yourself and thus _expected_ that I should do the same - which is very sensible, as I already wrote it to get my IDE to be helpful. If that is something you do not want _contributors_ to do, that would be a pertinent information to go into such a file.

Some project authors also [label issues](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels#about-default-labels) with labels such as 'good first issue' or 'help wanted', making it easier for contributors to find issues they are (not) supposed to touch.

If this would be more helpful than trying to fix things on my own, I also would not mind to create a Merge Request or two containing _only_ type annotations and true refactorings (behaviour-unchanging ones)?

Alexofp:
Yeah, separating stuff into different PRs is the best way probably. Issues fixes and refactorings should all be separate x3
And thank you again

Zsar:
Dunno, if that helps, but I always take care to segregate separate matters into different commits.

In my professional experience, most people do not bother to check commits, but as far as the tools are concerned, this would nominally be a great way to go about it.

So, somehow this is a "life hack" instead of just normal.

Zsar:
Tested for a while and found no adverse effects. => (hopefully) Ready

Zsar:
(Do you need/want me to do something else before Merge?)

Alexofp:
nah x3

Alexofp:
Thank youuuu

PR #323, opened 2026-05-18
[Fix] Missing Translated Buttons

PuroSlavKing:
When random "hypnotic" event uppears - translated buttons be gone, ill fix it

No translate:
image

Auto-translate result (Ru):
image
(Buttons silly gone)

### After patch:
image
(Buttons now here!)

It might cause problems somewhere else, but I quickly tested the localization outside of this event and everything was stable

UPD: It turned out that this is not the only place where the buttons disappear, so the fix can be considered global

### DIFF
--- a/Game/UI/GameUI.gd
+++ b/Game/UI/GameUI.gd
@@ -552,8 +552,8 @@ func translateText(manualButton = false):
 		var buttonsTexts = []
 		if(AutoTranslation.shouldTranslateButtons):
 			for optionID in options:
-				buttonsTexts.append(options[optionID][1])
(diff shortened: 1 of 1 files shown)


Alexofp:
I'm gonna merge this but I think there might be problems because some translators love to strip/replace random characters or add random spaces. I will have to test this.

Either way, thank youuu

Alexofp:
Tested it a bit. Seems to work pretty well in Google and Goggle batch translators. The other translators seem to be broken at the moment anyway so its fine x3. Thank you

PR #325, opened 2026-05-23
Update GlobalRegistry.gd [Android 16 Modded Fix]

kristijanWolff:
fixed the Android 16 not loading the modded version of the game. The issue was caused by Android's strict storage permissions (Scoped Storage), which block apps from directly accessing public directories like the "Downloads" folder. This caused the game to get stuck on an infinite loading screen.  The fix updates the paths in GlobalRegistry.gd to use Godot's internal "user://" directory instead. Because Android always permits an app to read its own sandboxed data folder, moving the custom mod files into "Android/data/com.viren.bdcc/files/BDCCMods/" allows the game to load them seamlessly without triggering any system security restrictions and it does this due to the moded version not need to reach to the documents folder. 
i used a Redmi mi 13 pro (Android 16 to test it).
! I USED A PHONE WITH HYPEROS,DONT KNOW IF SAMSUNG OR OTHER ANDROID DEVICES WORK !
also i used the 2.5.0 version as the version for the fix.

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -1,0 +1,50 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android" android:compileSdkVersion="35" android:compileSdkVersionCodename="15" android:installLocation="auto" package="org.rahimew.bdcc" platformBuildVersionCode="35" platformBuildVersionName="15">
+    <supports-screens android:largeScreens="true" android:normalScreens="true" android:smallScreens="true" android:xlargeScreens="true"/>
+    <uses-feature android:glEsVersion="0x20000" android:required="true"/>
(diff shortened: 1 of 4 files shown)


Alexofp:
Heyaa. Thank youu. Sadly, the whole point of using the documents folder was because the user:// folder is not reachable (on most androids?) normally unless you have root access or whatever. So there is not much I can do, it kinda has to be the way its currently is.

kristijanWolff:
Well im ordering a samsung A57, if it works itll cover most people i think, its not realy a fix but more of a workaround but if it helps at least one other person ill be happy.

kristijanWolff:
Oh,yeah i think it only fixes the HyperOS and maby a nother not 100% in line with google. Probably should rename it to "Android 16 HyperOS fix"

PR #336, opened 2026-06-06
check for restraints/collar when pulling towards tentacles, fix leash anim not being used

keerifox:
`"{punisher.name} clips a leash to {target.your} collar."` this line implies use of leash/collar which wouldn't work on guards/nurses/engineers unless they're wearing one

option A (this PR) - handle everything similar to stocks, i.e. require an equipped collar and use leash walking animation

option B - don't require collar for staff, then the message would be different, animation would ideally show them being pulled by the arm without use of a collar. would still need the restraint check like in this PR


Alexofp:
Yeah, handling it the same ways the stocks handle seems consistent. Thank youuuu

PR #339, opened 2026-06-12
fix ponytail 4 haircut being solid black and not supporting recoloring

keerifox:
haircut on a skulldog having solid black color only
 haircut on the same character now having 3 distinct colors


Alexofp:
Thankiesss

PR #340, opened 2026-06-12
add random name generator when starting new game or renaming owned pawns

keerifox:
i've seen too many comments from people being unable to try the game at all, due to the name input text field not working for them, this is sort of an alternate way to get past that input, and a feature for creatures that have difficulties deciding on a name for themselves or for someone else (even if they dont like the random names these might give them an idea for a custom one)

it's in a separate scene so can be reused for naming every single of your born kits or whatever

an option to choose a random name when asked to enter one

a random name Iris was generated. there is an option to choose between any, feminine or masculine name. then options to confirm or cancel


Alexofp:
Making it a separate scene might have been an overkill, it wouldn't take that much space if it was embedded directly into other scenes x3. But I guess it works.

Thank youuuu

keerifox:
oo possibly, it's only 2 scenes that are using it now, could be 3 or more in the future, i think having one place to make edits to is preferred in case someone wants to add robotic names to that etc!

PR #341, opened 2026-06-13
minimal changes

keerifox:
these are not entireeeely incorrect but yea "slave minimal level" sort of reads like "slave level barely anything", "minimum" would unambiguously mean it has to be "at least this number or more"


Alexofp:
Thankiesss

PR #342, opened 2026-06-13
fix runtime error when penis is rubbed against buttstacked subs

keerifox:
another pull request title that goes straight to my resume,, anyway

`getThroughClothingText` returns dom/dom1/sub/sub1 which need to be preprocessed and `addTextTopBottom` calls `addTextRaw` which explicitly does not, resulting in

You rub your cock against Clarence's backdoor  through !RUNTIME ERROR NO CHARACTER FOUND dom.yourHis []! uniform.

\-

feel free to make raw/notraw an argument of `addTextTopBottom`, or just directly call `processText`, i figured a separate function would be more optimized/readable but i'm fine with whichever way to fix this

there was also an unnecessary space symbol before "through"


Alexofp:
This whole thing is a mess but I dunno how to make it better, pain x3 Thank youuuuuuu

PR #344, opened 2026-06-27
allow mods to add new bodypart and inventory slots

Selinoccino:
- added ability to create and register new inventory and bodypart slots

seems to work quite well, even with stuff like datapack creation, but i haven't tested it much

could also change the vanilla slots to use this system, currently the custom ones are just extra

some screenshots:

screenshot of new "wings" bodypart slot appearing in datapack character creator)

screenshot of a new "piercings" inventory slot appearing in datapack character creation's default equipment section

screenshot of new wings appearing in their own section in the character creation scene

### DIFF
--- a/GlobalRegistry.gd
+++ b/GlobalRegistry.gd
@@ -224,6 +224,8 @@ var allSpecies: Dictionary = {}
 var items: Dictionary = {}
 var itemsRefs: Dictionary = {}
 var itemsByTag: Dictionary = {}
+var inventorySlots : Dictionary = {}
+var bodypartSlots : Dictionary = {}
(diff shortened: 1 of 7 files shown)


Alexofp:
Thank youuu. I will see if I can improve it a bit

Alexofp:
https://github.com/Alexofp/BDCC/commit/7cc5815819aad19d0a0f6d598022400561474d81

Improved some stuff. Fixed some bugs
Kept the list of slots for the MeshWithPattern. But now there is a CUSTOM entry in it and a bodypartSlotCustom variable to set, I think its a better design

PR #351, opened 2026-08-02
Improve undress chains and add npc fetish editor

Selinoccino:
for mods adding inventory slots that should prevent undressing items from other slots until being undressed themselves, made the undress chains be associated with a slot id, and a slot can only be undressed if all other equipped slots' undress chains have ended or the slot to be undressed is next in that chain

tldr items equipped "on top" of other items will never allow them to be undressed, until they're undressed themselves

my code for that is kinda crap but i tried my best :P

also added an npc fetishes editor for the devtools thing in the main menu, figured it fits there and would be handy

and also fixed a small issue in the character artwork panel that made it so having a null character and pressing the next button crashed the game

### DIFF
--- a/Game/SexEngine/Fetish.gd
+++ b/Game/SexEngine/Fetish.gd
@@ -43,6 +43,50 @@ const HypnosisSubject = "HypnosisSubject"
 const TFGiving = "TFGiving"
 const TFReceiving = "TFReceiving"
 
+static func getAllDefault() -> Array:
+	return [
+		VaginalSexReceiving,
+		AnalSexGiving,
+		AnalSexReceiving,
+		OralSexGiving,
+		OralSexReceiving,
+		Tribadism,
+		StraponSexVaginal,
+		StraponSexAnal,
+
+		Bodywritings,
+
+		Masochism,
+		Sadism,
+		Choking,
+
+		UnconsciousSex,
+		DrugUse,
+
+		Breeding,
+		BeingBred,
+		Condoms,
+		Lactation,
+		SeedMilking,
+
+		Exhibitionism,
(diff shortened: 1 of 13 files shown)


Alexofp:
It's better to separate things into different PRs x3 no need to shove the fetish editor into the same pr. Just for the future

Selinoccino:
a okay, thought i'd try not too spam too many prs ._.

Alexofp:
I'm not sure what this accomplishes. How are the undress chains different from the way they're currently are?

Alexofp:
I think I see the idea. I wanted only some items to be un-equipable. But I guess I can do it differently

Thank youuuu. I will do some changes, I hope its okay

Alexofp:
Did the changes here https://github.com/Alexofp/BDCC/commit/5d3363107ab72f7b46179966951bad5664f01dde
I dunno, I basically simplified everything I think x3
The chain is now just `slot: array of slots that need to be undressed first`
and it all works from there
added a variable to disable ability to undress a slot because now an empty array is a valid value (means you can always undress it)

Selinoccino:
yeah that looks faster than what i had :P

wouldn't adding something like a jacket to a mod still not work properly though? i think making the default undress chains a variable instead of a constant might be better, that way mods can add inventory slots that need to be free/unequipped before something like say underwear can be unequipped (that way maybe a single undress chain can be used instead, like "undress slot priority" or something)

Alexofp:
Making this any more complicated is not worth it, I'm sure there is only gonna be like 1 or 2 mods that add new slots, ahhhhhhhhhhhhhhhhhhhhh

PR #352, opened 2026-08-04
improve fetish generator

Selinoccino:
- added vaginal topping to getAllDefault function
- made default interest value editable for fetish generator

mostly an excuse to fix the first thingy, stupid oversight on my part ._.


Alexofp:
Thank youuuuuuuu

PR #354, opened 2026-08-04
add ability to view all mods by author in one click

keerifox:
the ability to search for `@fox` was already there (much secret, wow), this PR adds `"@keerifox"` for a case-sensitive exact match, and a published mods counter next to each listed mod author, that upon click filters the search to only display mods with that author listed

[2026-08-04_20-10-21.webm](https://github.com/user-attachments/assets/89f3dc32-3038-4f27-a560-43400dc8c028)

other tweaks/fixes:
\- enable X button within the search field to clear contents with one click
\- fix vertical scroll not being moved to the top when the search query changes
\- fix search query not being applied when changing the sort method

### DIFF
--- a/UI/ModBrowser/ModBrowser.gd
+++ b/UI/ModBrowser/ModBrowser.gd
@@ -151,7 +151,6 @@ func incrementModCountForAuthors(authorsData) -> void:
 func parseAuthorField(authorField:String) -> Dictionary:
 	var authors:Array = []
 	var preferredSeparator:String = ", "
-	authorField = str(authorField)
 	if(", " in authorField):
 		authors = authorField.split(", ")
 		preferredSeparator = ", "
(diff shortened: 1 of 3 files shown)


UI/ModBrowser/ModBrowser.gd
```diff
@@ -178,39 +227,66 @@ static func sort_oldest(a, b):
...
 func _on_CloseButton_pressed():
 	emit_signal("closePressed")
 
 
 func _on_RichTextLabel_meta_clicked(meta):
+	if(meta.begins_with(viewModsByAuthorUriPrefix)):
+		# moves caret to the end as opposed to using modSearch.text = "...
+		modSearch.clear()
+		modSearch.append_at_cursor("\"@"+meta.trim_prefix(viewModsByAuthorUriPrefix)+"\"")
+		updateModListWithSearchApplied(modSearch.text)
```
Alexofp: I'd do this like
```
setSearch(modSearch.text) # Or whatever function name you want
updateModList()
```

and then you make updateModList do all the fancy stuff that the updateModListWithSearchApplied currently does
what I learned over the years is that having a single UPDATE EVERYTHING function that doesn't take any arguments is very nice. With your approach there are now 2 UPDATE EVERYTHING functions but one of them is special. I dunno. I hope you understand what I mean x3

keerifox: o yea i can try that later

keerifox:
should be a little bit cleaner

Alexofp:
Looks good probably x3 thankiessssss

PR #359, opened 2026-08-12
switch 2 lines

Selinoccino:
switched 2 lines to add the ability for part3ds and stuff to reference their doll's character


Alexofp:
I hope it won't destroy everything x3 thankiessss

PR #360, opened 2026-08-19
moddable codeblocks

Selinoccino:
- add ability for mods to add their own codeblocks, but not replace default ones

very simple but makes it easier for mods, i know i wanna make some so :P
i saw that there was a sort of cache for them but figured i'd leave that alone, should be populated with them anyway, although maybe that's not desirable(?)

screenshot of a codeblock from a mod (duplicated the ishybrid one and changed the text)
image


Alexofp:
Seems good probably. I dunno, I think if people want to add custom codeblocks, they might as well pull request them into the game itself x3
Thankiess

PR #362, opened 2026-08-26
allow player to service dom without reputation, if sex work client explicitly asked for it

keerifox:
Lynne pleadingly gestures at you, wondering if you would have a different offer for her.. Lynne: 'O-Oh.. I'm not that good of a dom.. But.. m-maybe you would be..?' Looks like Lynne wants you to be a service dom.. Service Dom action description: You will be in charge. It takes more effort but your whore reputation is not high enough to demand more credits

was an oversight on my part that the option is greyed out at under 4 whore rep, specifically when a player offers the usual and the client requests a service dom instead.

the alternative is that they never ask until the player has enough reputation, but that doesn't seem fun,,

no 2x credits until the reputation level is reached however

### DIFF
--- a/Game/InteractionSystem/Interactions/Prostitution.gd
+++ b/Game/InteractionSystem/Interactions/Prostitution.gd
@@ -85,15 +85,16 @@ func found_client_text():
 	else:
 		addDisabledAction(ACTION_NAME_USUAL, ACTION_DESC_CLIENT_NOT_INTERESTED)
 	if(clientRequestedAskType in ["", "service"]):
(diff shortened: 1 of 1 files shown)


Alexofp:
Tired cat struggles to follow the logic but I trust you x3 thankiesss

keerifox:
aww it's basically

at 4+ whore rep:
can sub (1.0x credits earned)
can dom (2.0x credits earned)

at 1-3 whore rep:
can sub (1.0x credits earned)
can't dom, but if you offer to sub and they ask for dom instead, you can agree (1.0x credits earned)

without this pr you could not agree even when they asked, because reputation would be too low

Alexofp:
Thank you for explaining it <3

PR #364, opened 2026-08-30
fix extenders

Selinoccino:
- made all extenders load null data when starting a new game
- little thing to reload extenders when starting new game

very basic, mostly made this to fix the issue for a mod

### DIFF
--- a/Game/GameExtenders/GameExtender.gd
+++ b/Game/GameExtenders/GameExtender.gd
@@ -49,4 +49,6 @@ func saveData():
 	pass
 
 func loadData(_data):
-	pass
+	if _data==null:
+		get_script().reload()
+		return
--- a/Game/MainScene.gd
+++ b/Game/MainScene.gd
@@ -262,6 +262,9 @@ func startNewGame():
 		scene.queue_free()
 	sceneStack = []
 	
+	for gExtender in GlobalRegistry.getGameExtenders().values():
+		gExtender.loadData(null)
+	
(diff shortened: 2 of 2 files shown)


Game/GameExtenders/GameExtender.gd
```diff
@@ -49,4 +49,6 @@ func saveData():
 	pass
 
 func loadData(_data):
-	pass
+	if _data==null:
+		get_script().reload()
```
Alexofp, line 53: Extremely weird x3. Are you sure this the right way of doing this?

I will need more info about what the problem is

As is it currently, this line wouldn't get called anyway since extenders probably override the loadData function

Alexofp:
What mod? What issue? I don't wanna blindly merge this because I don't understand what it fixes, sorry, meow..

Game/GameExtenders/GameExtender.gd
Selinoccino: not really tbh, it's supposed to reset the state of the extender when the data to load is `null`, which should be the case when starting a new game

if you load a savefile and then start a new game from the main menu, the game extenders seem to carry their data from the loaded game onto the newly-started one, it's uh pretty weird ._.

and yeah, the extenders do override it, i was hoping modders would either do `.loadData(null)` or at least make their own version of resetting the state (maybe having a separate function would be better for that, which is called when starting a new game? i don't know :P)

PR #366, opened 2026-09-19
Relocate SexEngine.resetJustCame() call so that it isn't reset before SexActivities can access it

Mindstormsman:
Moves the SexEngine.resetJustCame() call out of SexEngine.playAnimation() and into GenericSexScene._run() so that it is reset after SexActivities have the opportunity to read it.

Currently _domInfo.justCame and _subInfo.justCame will always return false in a SexActivities getStartActions function.


Mindstormsman:
new commits from today:
added justCame to SexInfoBase.gd save data so that it properly persists through rollbacks

relocated SexEngine.resetJustCame call again,
it was initially placed to trigger at the end of the turn but this results in it being reset before the rollbacker can save it.
Now it is placed to trigger at the beginning of the turn so that anything that may need it has access to it and it is still reset before anything can operate on stale data from the previous turn.

